PR #253, opened 2015-08-06
Initialize missing fields in sensusUltraParse.m

sspagnol:
Uninitialized fields sample_data.dimensions and sample_data.variables cause Matlab (tested on R2014a) to throw an error, and hence the data file will not be imported.


ggalibert:
:+1:  @sspagnol, I will generalise this to other parsers.

PR #258, opened 2015-09-25
textscan BufSize option has been obsoleted

sspagnol:
At least from R2015b the BufSize option has been removed from textscan. Have chosen R2014b as a convenient cutoff (introduction of HG2). Don't know if this was important for performance in version <R2014b so have kept original call for <R2014b.

### DIFF
--- a/NetCDF/templateType.m
+++ b/NetCDF/templateType.m
@@ -141,7 +141,11 @@ function lines = readTemplate(filepath)
 
         if fid == -1, error(['could not open file ' filepath]); end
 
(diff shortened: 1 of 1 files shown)


sspagnol:
Also note that this doesn't make imos-toolbox fully HG2 compatible, just picking the obvious stuff first.

ggalibert:
Totally agree with your suggestion. However, in my opinion option BufSize was not necessary to parse template files (very small files...) in the first place so not using at all might be the best thing to do in order to fix this problem and keep the code simple. 

Can you please update your changes if you agree? I would then be happy to merge them.

sspagnol:
Hi Guillaume, hopefully a new PR has been created "textscan BufSize option
has been obsoleted #259"

Simon.

PR #259, opened 2015-09-29
textscan BufSize option has been obsoleted

sspagnol:
The BufSize option to textscan has been obsoleted in R2015b. As this function is reading in relatively small template files it should not be necessary for performance and so can be removed for all Matlab versions.


ggalibert:
:+1:

PR #263, opened 2015-10-19
Update executeDDBQuery.m

petejan:
Fix to read java.sql.Date classes, which is what is defined in the Java... schema for the database


ggalibert:
:+1:

PR #265, opened 2015-10-24
Flow Manager bug fix

BecCowley:
Change to FlowManager to use rawData (instead of nonUTCRawData) when applying any time PP routines to create autoQCData structure. This means that the time PP routines that are applied to the rawData are carried through to autoQCData. Previously, the timeOffsetPP routine reset the offsets to zero when run the second time.


BecCowley:
Sorry, will try that again, this time without all the other commits....

PR #266, opened 2015-10-24
Flow Manager bug fix

BecCowley:
Change FlowManager to use rawData (instead of nonUTCRawData) when applying any time PP routines to create autoQCData structure. This means that the time PP routines that are applied to the rawData are carried through to autoQCData. Previously, the timeOffsetPP routine reset the offsets to zero when run the second time.


ggalibert:
@BecCowley you should only include the following commits in this pull request :
f1b3fccb840a7f3c1487f41c0a8130b1a4892d55
3929fa2c04b853440f303d6cfcb18507a45c5a04

BecCowley:
@ggailbert, will do when I figure out how to. Should I close this pull request and try again?

PR #270, opened 2015-11-01
Allow TIME dimension to be plotted, even when the flag is set to zero.

ggalibert:


ggalibert:
Bec said : 
We need to have a discussion about the flags. Several parameters are not flagged as ‘good’ after selected QC routines are applied, therefore are not plotted. I don’t understand the logic of only applying a ‘bad’ flag, and not applying a ‘good’ flag to everything else when QC routines are applied. Can we talk about this next week?

ggalibert:
A couple of information in the meantime:

-There are two ways of plotting data from the menu (that is to say with vertical context, the same parameter measured by different instruments on the moorings are plotted):
1-plot everything without taking into account QC flags (plot non-QC'd data),
2-plot only the QC'd good data (plot good QC'd data, these are the default plots generated when exporting QC'd data in netCDF).

-imosInOutWaterQC is the only test that can flag part of a timeseries with good or bad while leaving another part non-QC'd. That's because we don't want to consider the data collected outside the time period the mooring is in position, but at this stage we cannot say that based on this single test the data collected inside this time period is good.

BecCowley:
The In out water test does not assign a 'good' flag. It only assigns a fail if it is out of the water and a 'raw' flag for in water periods. In what routine do we assign a 'good' flag to the time variable?

ggalibert:
In imosImpossibleDateQC test: https://github.com/aodn/imos-toolbox/wiki/QCProcedures#impossible-date-test---imosimpossibledateqc---compulsory

bhasin85:
@ggalibert Can we close/merge this PR?

PR #272, opened 2015-11-06
replace call to de2bi (which exists only in Communication Toolbox)

sspagnol:
de2bi only exists in Communications Systems Toolbox. Replacement code needs to be checked it performs similarly (I can't as I don't have the Communications Systems Toolbox).

### DIFF
--- a/Parser/readParadoppBinary.m
+++ b/Parser/readParadoppBinary.m
@@ -390,7 +390,7 @@ sect.Time        = readClockData(data, idx+4);
 block2           = data(idx+10:idx+23); % int16
 
 sect.PressureMSB = data(idx+24); % uint8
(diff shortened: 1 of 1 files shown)


Parser/readParadoppBinary.m
```diff
@@ -390,7 +390,7 @@
 block2           = data(idx+10:idx+23); % int16
 
 sect.PressureMSB = data(idx+24); % uint8
-sect.Status      = uint8(de2bi(data(idx+25), 8))'; % 8 bits status code http://cs.nortek.no/scripts/customer.fcgi?_sf=0&custSessionKey=&customerLang=en&noCookies=true&action=viewKbEntry&id=7
+sect.Status      = uint8(dec2bin(data(idx+25), 8)-'0')'; % 8 bits status code http://cs.nortek.no/scripts/customer.fcgi?_sf=0&custSessionKey=&customerLang=en&noCookies=true&action=viewKbEntry&id=7
```
ggalibert: This is not equivalent unfortunately.

ggalibert: uint8(flipud(str2num(dec2bin(data(idx+25), 8)'))) should be.

PR #274, opened 2015-11-09
Remove hardwired TIME dimension id access.

sspagnol:
Some of the graphing routines are grabbing the TIME dimension id directly from the variables dimension layout, is there a reason for this? Assuming there isn't below set of patches uses the standard getVar(sample_data.dimensions, 'TIME') syntax.

If this is ok, then there are cases of routines looking for frequency or direction related dimension eg flagTimeSeriesTimeFrequency.m that could be updated in a similar fashion.


ggalibert:
The case with TIME is unambiguous, so yes there's no reason not to apply your patch.

However with depth, this is actually not DEPTH, but HEIGHT_ABOVE_SENSOR or DIST_ALONG_BEAMS depending on cases so we need to implement a smarter logic. Referring to 2 was just straightforward although it is agreed not elegant.

For freq, it can actually be FREQUENCY or FREQUENCY_2. Again we would need a smarter logic.

For dir it might not be a problem to replace it by the relevant getVar call.

PR #277, opened 2015-11-10
replace call to de2bi (which exists only in Communication Toolbox)

ggalibert:


ggalibert:
@sspagnol would you be happy with this?

sspagnol:
If that matches what you need then all good for me.

PR #278, opened 2015-11-11
Resize buttons

petejan:
pull request with only changes for this feature


ggalibert:
There shouldn't be any modification in NetCDF/template/global_attributes_timeSeries.txt for this PR.
Two distinct commits doesn't make sense here, one single commit is neater.

I won't merge these changes because of the resulting look on Linux. Changes on GUI must be tested on Linux and Windows platforms at least.

ggalibert:
re-size_buttons

PR #280, opened 2015-11-11
changes for generic database

petejan:
remove java/...../schema classes, and get field names from database
result set, not from classes.
Also include Java code as a eclipse project to make maintenance easier.


ggalibert:
There is too much going on in one single commit to make an easy review.
One commit for update to latest UCanAccess version, one for getting field names from database and one for better Eclipse project maintenance would be a minimum.

Java/src/org/imos/ddb/JDBCDDB.java
```diff
@@ -33,160 +33,193 @@
...
-	  this.user = user;
-	  this.password = password;
-	  
-	  // Test connection - throws exception at creation if can't connect
-      Class.forName(driver);
-      
-      Connection conn = DriverManager.getConnection(connection, user, password);
-      conn.close();
-  }
+public class JDBCDDB extends DDB
```
ggalibert: The java code reformatted on top of your personal changes for a generic database schema changes makes it very hard to highlight your personal changes to the original code...

petejan:
I like to keep the commits atomic. There is no need to update the ucacaccess to v3 as the current version works just fine. The commits for database updates need to have the updates to ucancess because of case issues. The eclipse project files just make life easier.

Java/src/org/imos/ddb/JDBCDDB.java
petejan: The code was so ungly to read I could not figure out what was going on.Once this PR is merged the code will be much more readable. 
Any technical issues with this way of dealing with the database tables?
Does the code work for you on your version of the database?

Java/src/org/imos/ddb/JDBCDDB.java
```diff
@@ -33,160 +33,193 @@
...
+
+				if (fieldValue == null)
+					throw new Exception("a fieldValue must be provided");
+
+				// wrap strings in quotes
+				if (fieldValue instanceof String)
+					query += " WHERE \"" + fieldName + "\" = '" + fieldValue + "'";
+				else
+					query += " WHERE \"" + fieldName + "\" = " + fieldValue;
+			}
```
petejan: Changes here to fix case of field values by quoting the strings and field names

Java/src/org/imos/ddb/JDBCDDB.java
```diff
@@ -33,160 +33,193 @@
...
+
+				results.add(instance);
+
+				for (int i = 1; i < columnsNumber; i++)
+				{
+					DBObject db = new DBObject();
+
+					db.name = rsmd.getColumnName(i);
+					db.o = rs.getObject(i);
+
```
petejan: Create a DBObject with the name (from the column name) and the object from the database

Java/src/org/imos/ddb/Test.java
```diff
@@ -30,111 +30,175 @@
...
+		//System.out.println("args " + args.length);
+		
+		if (args.length > 0)
+		{
+			if (args.length < 2)
+			{					
+				System.err.println("example usage : org.imos.ddb.Test net.ucanaccess.jdbc.UcanaccessDriver \"jdbc:ucanaccess://C:/Users/jan079/svn/imos-toolbox/TAS-IMOS_DeploymentDatabase2.0.mdb\"");
+				System.err.println("                org.imos.ddb.Test org.postgresql.Driver \"jdbc:postgresql://localhost/IMOS-DEPLOY\" jan079");
+				
+				System.exit(-1);
```
petejan: Add possible args to the Test class to send driver/connection url from the command line at execution time

NetCDF/template/global_attributes_timeSeries.txt
```diff
@@ -41,8 +41,8 @@ N, geospatial_lat_max       = [ddb Site Sites Latitude]
 N, geospatial_lon_min       = [ddb Site Sites Longitude]
 N, geospatial_lon_max       = [ddb Site Sites Longitude]
 N, instrument_nominal_height= 
-N, instrument_nominal_depth = 
-N, site_nominal_depth       = 
+N, instrument_nominal_depth = [ddb InstrumentDepth]
+N, site_nominal_depth       = [ddb Site Sites Depth]
 N, site_depth_at_deployment =
```
petejan: Added these as the default

Java/src/org/imos/ddb/ODBCDDB.java
```diff
@@ -38,146 +38,171 @@
...
+				results.add(instance);
+
+				Field[] fields = clazz.getDeclaredFields();
+
+				// set the fields of the object from the row data
+				for (Field f : fields)
+				{
+					DBObject db = new DBObject();
+					db.name = f.getName();
+
```
petejan: Changes to return an array of DBObject objects the same as the JDBC interface

Java/src/org/imos/ddb/JDBCDDB.java
```diff
@@ -33,160 +33,193 @@
...
+			String query = "SELECT * FROM " + tableName;
+			if (fieldName != null)
+			{
+
+				if (fieldValue == null)
+					throw new Exception("a fieldValue must be provided");
+
+				// wrap strings in quotes
+				if (fieldValue instanceof String)
+					query += " WHERE \"" + fieldName + "\" = '" + fieldValue + "'";
```
ggalibert: Are the double quotes for fieldName necessary for you? It doesn't work for me with them.

Java/src/org/imos/ddb/JDBCDDB.java
ggalibert: Then a single commit with just reformatting and no modification would be clearer to me before making any further modification to the code.
Issues for me were the double quotes on the field names. I had to remove them for the toolbox to work.
Bouncy Castle needs to be no more recent than 1.50. Jackcess encrypt is necessary for the AIMS database and it doesn't support Bouncy Castle from 1.51.

ggalibert: Not sure I want to add Eclipse project files to the repo. They probably change from platform/config/version to another and not everyone is using Eclipse.
At the moment anyone with a JAVA_HOME environment variable set can compile the code by doing ```ant install``` in the Java directory.

NetCDF/template/global_attributes_timeSeries.txt
ggalibert: I'm not comfortable to do that since these are not valid config for everyone at the moment.

Java/src/org/imos/ddb/JDBCDDB.java
petejan: Hmm, what does not work, does it not find the fieldnames, maybe an update to the .txt files so the names match the schema in the database.
Use the Test class to see what the fieldnames case is in the database.

PR #282, opened 2015-11-16
refactor geomag code

petejan:
refactor geomagnetic declination code, split from PreProcessor,
Add macosx calculation binary
Add (currently not used) matlab implementation of geomag calculation,
also add dialog to prompt for lat/lon/depth if not available from database


ggalibert:
Could you please remove commits de2c867 and cd514fb from this branch?

Also, the second commit 04cda01 can be squash/fixup with the first one 1dc0cc8 (see git rebase).

ggalibert:
In the first commit, I don't think it is relevant to add the C source code yet. Better in https://github.com/aodn/imos-toolbox/pull/283.

petejan:
The bug fixed in commit 04cda01 was in the original magneticDeclinationPP.m code as well if generating more than one declination, which is where I copied the code from. I only found it later when testing.
I added the C source as I build the OSX binary from it, in the first commit. 
Do you only want one commit in the pull request?

ggalibert:
I have already started to clean/modify this branch. Also noticed a couple of bugs including the fact that windows exe only supports IGRF files with CRLF line termination while linux/mac exe only support LF line termination... Will push in a different branch when ready.

petejan:
If the CRLF/LF problems in the coefficient file or just the input file? I thought the geomag70.m had this code in it already.
Why not push to the existing pull request/branch?

ggalibert:
The line termination problem is in the .COF files. At this stage of the compiled C code, it looks like we need both versions of the .COF.
I cannot push on a branch that is on your repo petejan. New PR is https://github.com/aodn/imos-toolbox/pull/286

PR #288, opened 2015-11-19
Handle most raw Nortek  binary structures and more processed Storm structures

sspagnol:
Read in structures
Id54 readAwacWaveDataSUV % 0x36
Id66 readAwacStageData % 0x42
Id96 readWaveParameterEstimates % 0x60
Id97 readWaveBandEstimates % 0x61
Id98 readWaveEnergySpectrum % 0x62
Id99 readWaveFourierCoefficentSpectrum % 0x63

Modify readParadoppBinary to correctly handle some special Id's with variable size fields, currently only two (Id66 and Id98).

Standardize field names, some where from C struct names which didn't match the style of the rest of the code.

The only structures not handled are
Id34 (0x22) 2-beam aquadopp
Id35 (0x23) 1-beam aquadopp
Id37 (0x25) 2-beam continental
Id38 (0x26) 1-beam continental
Id7 (0x07) vector probe check

To be confirmed of Id96-Id99 : Data that is determined to be invalid in the wave processing is flagged with a hex value of 0xFFFF. This means that the value for unsigned is 65636 and it is -32768 for signed
values.

### DIFF
--- a/Parser/readParadoppBinary.m
+++ b/Parser/readParadoppBinary.m
@@ -95,8 +95,10 @@ structures = struct;
 
 % list of Ids that have variables that can vary in size, so cannot just
 % concatenate them
-specialIds{1}.id = 'Id98';
-specialIds{1}.vars = { 'Energy' };
+specialIds{1}.id = 'Id66';
+specialIds{1}.vars = { 'Amp' };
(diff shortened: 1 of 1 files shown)


ggalibert:
:+1:

PR #290, opened 2015-11-26
Geomag fix path max length

ggalibert:


ggalibert:
@petejan this is how I would have liked to see the PR with different commits to be able to see the changes you introduced to geomag.c in the second commit.

By the way, I compiled your modified C code with the following command:
```
gcc geomag70.c -o linux/geomag70 -lm
```
It doesn't complain but then when I execute the binary I get a segmentation fault error. Could you please have a look at what could be wrong?

PR #291, opened 2015-11-26
commit OS X and linux binaries

petejan:
build OS X, and linux binary for i386


ggalibert:
What's the geomag.c source code associated with these?

I've built win32 (pei-i386), linux64 (elf64-x86-64) and macosx (x86-64, what's yours?).

petejan:
I downloaed the geomag.c source from your repository.

gdb is the only way to find the segfault error.


Compile with the –g flag eg



gcc geomag70.c –g -o linux/geomag70 -lm

And then run gdb linux/geomag70

I complied the linux version for i386, its quite tricky to create binaries for all these platforms, which is why I pulled in a matlab implementation of the geomag code.

Pete

PR #292, opened 2015-11-30
geomag is able to read files with CRLF or LF endings for any architec…

ggalibert:
…ture.


ggalibert:
@petejan I included the modification on source C code from this commit https://github.com/aodn/imos-toolbox/commit/9e2e9ff024c8f7c5c2fb009773226259c6229a3f to this PR.

PR #293, opened 2015-12-03
Fix hard coded ddb fields

ggalibert:


ggalibert:
@petejan these are the changes I'm ok to implement so far. Fell free to let me know if that works on your side.

petejan:
Incremental changes in branch,

generic-db2

PR #295, opened 2015-12-03
Salinity from TEOS10 consistently calculated, even for dataset that a…

ggalibert:
…lready include salinity


ggalibert:
@sspagnol what do you think of this, could that be a good idea?

Instruments/Manufacturers can still output salinity using EOS-8 while the toolbox is calculating it with TEOS-10, but only when it doesn't exist already. So for now we end up with an inconsistent collection of salinity. Is this a big deal?

I've been reported:
```Ruskin uses the UNESCO-80 formulae for calculation of salinity (and other quantities), giving salinity on the Practical Salinity Scale (PSS). We have talked about implementing TEOS-10 in Ruskin, but currently don't have any plans for it. The reason for this is that Absolute Salinity is calculated from Practical Salinity, and PS is still the quantity that should be archived.```

petejan:
Don't think this is a very good idea. 
Practical salinity is what should be calculated and stored, Absolute salinity should NOT be calculated and stored when measured using conductivity. see http://teos-10.org/ 
"Importantly, while Absolute Salinity (g/kg) is the salinity variable that is needed in order to calculate density and other seawater properties, the salinity which should be archived in national data bases continues to be the measured salinity variable, Practical Salinity (PSS-78)"

The formula in TEOS-10 for practical salinity and EOS-80 is the same, based on PSS-78.

sspagnol:
My understanding was the toolbox only calculated (when needed) and stored
practical salinity using TEOS10 gsw_SP_from_R which returns SP = Practical
Salinity on the PSS-78 scale. As Peter stated "The formula in TEOS-10 for
practical salinity and EOS-80 is the same, based on PSS-78." so I assumed
all was ok. And it sounds like even if Ruskin implement TEOS10 they will
still output SP.

No in terms on whether you want to track where SP was calculated
(manufactures s/w or toolbox) I don't have any ideas of pros/cons.


For what is worth I have always told people here not to output SP from
instrument s/w (just required cond/temp/pres) and let the toolbox calculate
it.

Simon.

petejan:
Hmm, now sure that this is the best way. 

The instruments are calibrated in Practical Salinity (yes using conductivity as the transfer standard), using the manufactures formula. The Salinity is considered the measured value, not conductivity, any faults/rounding in the manufactures implementation of PSS-788 should be taken up in conductivity to give the correct calibrated Practical Salinity.

ggalibert:
Ok, thanks, I wasn't sure whether formulas to calculate practical salinity in EOS-8 and TEOS-10 were the same.

Now this pull request is just about systematically calculating PSAL with TEOS-10 rather than just calculating it when the dataset doesn't already include PSAL from the instrument. The instrument PSAL would still be left intact in the dataset.

PR #299, opened 2015-12-15
Fix paradopp

ggalibert:
Most of the time:
-battery and analog inputs are uint16,
-temperature is int16 (and can now support negative values...),
-status and error are int8 (error can sometimes be int16).

Checksum is left uint16 (although defined int16) because of how the check is implemented in Matlab. Retrieved value doesn't matter much, what matters is that the check finds consistent values.


ggalibert:
@sspagnol , made some improvements to readParadoppBinary if you want to have a look before I merge.

sspagnol:
Looks ok. Not in the office until next year so can't test. If it reads in the example file you have, then all good to go.

PR #304, opened 2016-01-28
Handle case where times from spectra are shorter than LOG9.txt wave data times.

sspagnol:
Have had a case where instrument problems create multiple PD0 files and after concatenating/renumbering and processing the resultant spectra times are shorter than LOG9.txt wave data times. Typically missing the last record, but to cover all possibilities copy data from spectra times to matched LOG9 time.

I have never had a case where spectra times are longerthan LOG9.txt wave data times so just issue a warning. If this should occur then will need to revisit that part of the code.

And it looks like my Matlab smart indent has created some whitespace shifts, sorry, don't know how to create diff against master branch ignoring this.

### DIFF
--- a/Parser/readWorkhorseWaveAscii.m
+++ b/Parser/readWorkhorseWaveAscii.m
@@ -1,21 +1,21 @@
 function waveData = readWorkhorseWaveAscii( filename )
-%READWORKHORSEWAVEASCII Reads RDI Workhorse wave data from processed wave text files 
+%READWORKHORSEWAVEASCII Reads RDI Workhorse wave data from processed wave text files
 % (_LOG9.txt, DSpec*.txt, PSpec*.txt, SSpec*.txt and VSpec*.txt).
 %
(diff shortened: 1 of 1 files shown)


Parser/readWorkhorseWaveAscii.m
```diff
@@ -201,4 +201,39 @@
...
+        oldTime = waveData.(specName).time;
+        sizeOldData = size(oldData);
+        waveData.(specName).time = waveData.param.time;
+        waveData.(specName).data = NaN([nTimes sizeOldData(2:end)]);
+        newdataLoc = repmat(loc, [1 sizeOldData(2:end)]);
+        waveData.(specName).data(newdataLoc) = oldData;
+        clear('oldData');
+        clear('oldTime');
+        clear('newdataLoc');
+    else
```
ggalibert: This line needs to be ```elseif nSpecTime > nTimes``` instead of ```else```since nSpecTime == nTimes is not a problem.

ggalibert:
Closing this PR for this new one https://github.com/aodn/imos-toolbox/pull/305

PR #305, opened 2016-01-28
Handle case where times from spectra are different (currently always …

ggalibert:
…shorter) than LOG9.txt wave data times.


ggalibert:
sspagnol commented:



Have had a case where instrument problems create multiple PD0 files and after concatenating/renumbering and processing the resultant spectra times are shorter than LOG9.txt wave data times. Typically missing the last record, but to cover all possibilities copy data from spectra times to matched LOG9 time.

I have never had a case where spectra times are longerthan LOG9.txt wave data times so just issue a warning. If this should occur then will need to revisit that part of the code.

And it looks like my Matlab smart indent has created some whitespace shifts, sorry, don't know how to create diff against master branch ignoring this.

PR #311, opened 2016-02-13
Additional WQM turbidity string

sspagnol:
Newer WQM firmware have another turbidity string. Applies to 2.5 as well.


ggalibert:
Included in https://github.com/aodn/imos-toolbox/pull/314 to deliver into version 2.5.2.

PR #312, opened 2016-02-15
Open geomag output file in text mode

sspagnol:
Need to open goemag output file in text mode to allow Matlab to handle unix \n and dos \r\n correctly. If you don't then textscan only reads the first line.


ggalibert:
Included in https://github.com/aodn/imos-toolbox/pull/314 to deliver into version 2.5.2.

PR #318, opened 2016-02-19
Missing 'mode' string in call to SBE3x

sspagnol:


ggalibert:
Thanks, will include this in a new PR with updated binaries for version 2.5.4

PR #321, opened 2016-02-24
XR420 parser header.correction field already handled 

sspagnol:
header.correction field already correctly handled a few lines below at the first if statement (old line 79, new line 78)


ggalibert:
Thanks @sspagnol ,

will include this in a new PR with updated binaries for version 2.5.4

PR #328, opened 2016-03-16
Avoid spurious warning in fileRemoveCallback

sspagnol:
Avoid warning generated if user tries to delete a file from an empty deployment file list


ggalibert:
Thanks, will include that in 2.5.5 via separate branch (Sorry, have to do so for anything that can go into 2.5 branch so that I have binaries coming at the same time as the code modifications).

PR #345, opened 2016-04-28
Handle older SBE37SM header style

sspagnol:
Have some older style SBE37SM cnv files which have a header line like

'* SBE37-SM V 2.6b  SERIAL NO. 4481    10 Feb 2014  04:33:43'

Current code didn't  pull out instrument_model, instrument_firmware, instrument_serial_no


ggalibert:
Thanks for this. Will include it in the next version along with updated binaires.

ggalibert:
Included in https://github.com/aodn/imos-toolbox/pull/347/commits/98b627422b6695be7bfb157db696f62563cfd61a

PR #359, opened 2016-06-14
Nortek parser to skip bad sync

petejan:
At the moment the Nortek pasers aborts when a bad checksum and bad sync are found. This truncates the output file. 

This change simply keeps scanning for sync characters until the next sync/good check sum


ggalibert:
Thanks! Will include that in a next version as soon as possible. In the meantime I wonder if you re-play the file with Nortek software it might be able to "clean" such corrupted file once and for all.

petejan:
Hmm the Nortek software seems to load the file fine, and reports 1 checksum, but I don’t know if you can re-save the file without the bad checksum.

ggalibert:
Once loaded you should be able to do an export.

ggalibert:
Ok maybe not with the Nortek free softwares but maybe with Storm?

ggalibert:
Fixed in v2.5.13 by commit https://github.com/aodn/imos-toolbox/commit/17dfe9b96e574be6a1bdec5a5ec4050a4ed05e3c

PR #364, opened 2016-07-20
New TEMP variable name for RBR parser

sspagnol:
Added new TEMP string for RBR solo instrument. While doing this aslo had issue with logic to construct format string to read in data. Code tidy to split variable names line (strreps -> regexprep) then fix logic to construct format string (length(cols)-1 -> length(cols)-2), you already have two known '%s' format cols (Date and Time).


ggalibert:
Fixed in https://github.com/aodn/imos-toolbox/commit/741c95d1212df4e71a65703ce61c193b483a3cb4

ggalibert:
See v2.5.14

PR #365, opened 2016-07-20
Change logic to handle SBE16 cnv file with mutliple hdr lines

sspagnol:
Try to handle SBE cnv files that have multiple hdr|cast date lines for profile and timeSeries data, these dates are used to try and calculate time of first data point. Assume that if you are in profile mode then you require all the hdr|cast lines, else you are in timeSeries mode then only need the first occurence.

Issue arose with an SBE16 cnv file (moored instrument) that had multiple hdr date lines, so that header.castDate ended up being an array which caused problems in convertSBEcnvVar when calculating TIME.

This worked on moored SBE16 timeSeries file with multiple hdr line, one without a hdr line and a CTD profile file. Any other testing ideas?

Added mode string to both functions parseInstrumentHeader and parseProcessedHeader but is only current used in parseInstrumentHeader.


ggalibert:
Do you really want to add support for multiple hdr date lines in profile mode? If you follow Tim Ingleton's pre-processing documentation, there should be only one time stamp in the profile.

sspagnol:
Trying to handle SBE16 moored on NRSDAR, they have the following lines in the header

* <Headers>
* hdr   1 05 Aug 2015 04:44:59 samples 1 to 2000, int = 900, stop = logging
* hdr   2 26 Aug 2015 00:45:14 samples 2001 to 4000, int = 900, stop = logging
* hdr   3 15 Sep 2015 20:45:14 samples 4001 to 6000, int = 900, stop = logging
* hdr   4 06 Oct 2015 16:45:14 samples 6001 to 8000, int = 900, stop = logging
* hdr   5 27 Oct 2015 12:45:14 samples 8001 to 10000, int = 900, stop = logging
* hdr   6 17 Nov 2015 08:45:14 samples 10001 to 12000, int = 900, stop = logging
* hdr   7 08 Dec 2015 04:45:14 samples 12001 to 14000, int = 900, stop = logging
* hdr   8 29 Dec 2015 00:45:14 samples 14001 to 16000, int = 900, stop = logging
* hdr   9 18 Jan 2016 20:45:14 samples 16001 to 17846, int = 900, stop = stop cmd

So just needed some way of handling that. Ideally the user would have exported with timeK/timeY format.

ggalibert:
Fixed in https://github.com/aodn/imos-toolbox/commit/c55915836fd337a03120150a233235e438e70c2a

ggalibert:
See v2.5.14

PR #369, opened 2016-08-10
Handle change in case of variable names for new version of WQMHost in readWQMdat.m

sspagnol:
Newer WQM firmware require newer version of WQMHost to convert the raw file to a dat file. There has been a change in case for some variables e.g. version 1.27  'Turbidity(NTU)' -> version 1.59 'TURBIDITY(NTU)'). To work around this all tokens have been defined in upper case, and the variable name header line is converted to upper case when read in. All references to variables (e.g. DO conversion lines 300-333) are made against upper case names.

So far looks like only turbdity and chlorophyll names changed. This PR was the cleanest way I could think of to handle this issue with some future proofing (hopefully).

An short example of a newer WQMHost v1.59 DAT file header is below

WQM SN:  119
File Name:  WQM0119.003
Created On:  020216	001221
Control:  Autonomous
Sample Seconds:  60
Sample Interval Seconds:  900
Data Averaging:  None

WQM	WQM-SN	MM/DD/YY	HH:MM:SS	Volts(VDC)	Cond(S/m)	Temp(C)	Pres(dbar)	Sal(PSU)	DO(ml/l)	CHLA(UG/L)	TURBIDITY(NTU)
WQM	119	02/02/16	00:14:27	15.63	0.00001	26.9750	0.00	0.013	5.784	-0.178	0.510
WQM	119	02/02/16	00:14:28	15.62	0.00001	26.9744	0.00	0.013	5.783	-0.178	0.510
WQM	119	02/02/16	00:14:29	15.63	0.00001	26.9747	0.00	0.013	5.784	-0.178	0.510

### DIFF
--- a/Parser/readWQMdat.m
+++ b/Parser/readWQMdat.m
@@ -117,6 +117,11 @@ function sample_data = readWQMdat( filename, mode )
   % in the input file) to the IMOS compliant parameter name. It also contains 
   % comments for some parameters.
   %
+  % Note that different versions of firmware require different versions of
+  % WQMhost to convert the raw file to a dat file. Some of the headers have
+  % different case. To work around this all tokens are defined upper case 
+  % and header line is converted to upper case when read in
+  %
   params{end+1} = {'WQM',                   {'',     ''}};
   params{end+1} = {'SN',                    {'',     ''}};
   params{end+1} = {'WQM-SN',                {'',     ''}};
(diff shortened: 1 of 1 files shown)


ggalibert:
Thanks!
An equivalent work-around is implemented in commit https://github.com/aodn/imos-toolbox/commit/e9f69525138408b9641a90f672140885c6f3a999 which has been included in https://github.com/aodn/imos-toolbox/releases/tag/v2.5.15

PR #372, opened 2016-08-23
Keyboard shortcuts to pan/zoom data in main display

sspagnol:
Keyboard shortcuts to zoom/pan on current axis (which is sometimes more convenient than going to toolbar and selecting the requisite tool).

Keys you can use are:
z, Z: zoom in, zoom out, in both dimensions
x, X: zoom in, zoom out, x dimension only (for all plots)
y, Y: zoom in, zoom out, y dimension only (only current selected plot)
arrow keys: pan the data (left/right all plots, up/down only current selected plot), shift arrow increase panning factor
a: axis auto (reset)

With multiple plots in main window the plot must first get focus by user selecting it with left-mouse-click. This PR does not change current behaviour for a 2D plot (eg UCUR), left-mouse-click still produces the a profile plot.


ggalibert:
I'm definitely in favour of making easier scrolling through a dataset when zoomed in. I've just tried these shortcuts and I'm not sure this is the easiest way to achieve painless scrolling through a dataset. I was imagining the following solution that I will try to implement in the future:
-When manually zooming with the mouse, a horizontal slider appears at the bottom of the GUI to enable the user to scroll along the X axis.

taking into account that:
-Displayed 1D variables must have a consistent X axis and stay sync'ed.
-Displayed 2D variables must have a consistent X and Y axis and stay sync'ed.
-Displayed 1D variable could have their Y axis automatically updated to adapt to the variation of this variable.

sspagnol:
How do you imagine the scrolling (through time) to work? If you have to drag a bar across then I would find using left/right arrow keys much more easier. And a slider would just eat up more screen realestate.

Regards comments
-Displayed 1D variables must have a consistent X axis and stay sync'ed.
Agree.

-Displayed 2D variables must have a consistent X and Y axis and stay sync'ed.
Agree.

-Displayed 1D variable could have their Y axis automatically updated to adapt to the variation of this variable.
Disagree. Personally like being able to control Y axis so can get a feel for rate-of-change of a variable, and if the y-scale keeps jumping around then that could be annoying.

ggalibert:
How about having special keys like "Ctrl" or "Shift" that would change the "left click and drag" pointer behaviour?
Left click and drag" behaviour could be:
-Default => select data to manually flag.
-With "Ctrl" key => pan.
-With "Shift" key => zoom in.

Right click on a flagged data shows the flag meaning, but outside a data could should the zoom/pan context menu and in particular the link to reset to original view.

sspagnol:
How do you envisionage the "Ctrl" key pan to work? Would a plot display be divided into 4 regions diagonally (upper left corner to lower right corner, and upper right cornet to lower left corner) then depending on where you click you pan in that direction e.g. click in north region pan up, click in the west region pan left?

If have "Shift" to zoom in then would also need a key to zoom out.

I liked the keyboard only shortcuts idea as it would operate independent of what mode you where in. The only thing I couldn't work out was having to click in the plot display to get focus.

ggalibert:
I kind of like using the mouse and click to do all of that except I have to travel back and forth to the zoom / pan icon on the upper left corner of the window to enable / disable the relevant mode. 

I prefer zoom / pan via the mouse as I feel I have a better control and understanding of what's going on. I can zoom in exactly where I want and / or pan the exact amount I want.

I was hoping I could use "Shift" or "Ctrl" keys to change the default "click and drag" action which is select points for manual QC to panning on the dataset (like when you click and drag on Google maps) or zooming in on the dataset (like when you click and drag with the zoom enabled on a Matlab figure). The zoom out/reset actions would be available via the context menu displayed by a right click.

Having a quick go at this in Matlab looks more complex than I thought though...

ggalibert:
Pretty keen on implementing this to start with: https://au.mathworks.com/matlabcentral/answers/169095-are-there-hot-keys-shortcuts-that-enable-interactive-mouse-controlled-pan-and-zoom-for-a-figure

ggalibert:
In GUI/mainWindow.m line 321, add:
```Matlab
  hHotKeyMenu = uimenu(fig, 'label', 'Hot Keys');
  uimenu(hHotKeyMenu, 'Label', 'Enable zoom',       'Accelerator', 'z', 'Callback', @(src,evt)zoom(fig, 'on'));
  uimenu(hHotKeyMenu, 'Label', 'Enable pan',        'Accelerator', 'a', 'Callback', @(src,evt)pan( fig, 'on'));
  uimenu(hHotKeyMenu, 'Label', 'Disable zoom/pan',  'Accelerator', 'q', 'Callback', @disableZoomAndPan);
```
And then somewhere add:
```Matlab
function disableZoomAndPan(source, ev)
    pan(fig, 'off');
    zoom(fig, 'off');
end
```

ggalibert:
This is the hot key implementation I suggested above: https://github.com/aodn/imos-toolbox/commit/633c5abf7000eadbcc7e3ad00bf582bb77b44d04 in https://github.com/aodn/imos-toolbox/releases/tag/v2.5.20

bhasin85:
@ggalibert Can we close/merge this PR?

PR #392, opened 2016-11-21
Promote both xdata/ydata to type double in pcolor call for time-varia…

sspagnol:
…ble plots. If you pass in type single for ydata to pcolor then both axis data types become single, which caused problems in R2016b for datestr trying to convert type single to a string.


ggalibert:
Thanks for that, will include in the next version of the toolbox.

ggalibert:
Included in https://github.com/aodn/imos-toolbox/commit/58f2279668d0bc1cc77066551f9a08c3f2a235b8 in v2.5.21

PR #397, opened 2016-11-23
parser for Aanderaa RCM meter text file

petejan:
Parses output from Aanderaa software for old RCM-8 and SeaGuard instruments


ggalibert:
Thanks for that, will include this new parser in next version of the toolbox.

ggalibert:
In the [imosParameters.txt](https://github.com/aodn/imos-toolbox/blob/master/IMOS/imosParameters.txt), there is no such parameter as:
-STD
-ABSI

What that means is these variables will be read from the instrument files and labelled STD or ABSI (without any definition nor unit) but they will not be output by the toolbox.

ggalibert:
@petejan are you happy with this situation or would you like those parameters to be written in the NetCDF file?
If you would like the later option then I would need a definition of what these parameters are with their unit.

* STD or "Single-Ping Standard deviation": I imagine this to be the standard deviation of current speed (would need to be converted to m.s-1) measured by a transducer at the ping level. I imagine this would be on average for all 4 transducers.
* ABSI or "Signal strength": following definitions from ABSI1, could be something like the backscatter intensity from acoustic beam but again I imagine this would be on average for all 4 transducers.

If you think we need to store these parameters then can you please confirm / correct their definitions?

petejan:
Been meaning to dig into these,

ABSI is the signal strength in dB, so need to add a line to imosParameter.txt for it.
The STD, is the Statistic variance (std) of the velocity measurment, in cm/s, should add the conversion to the file reader.

If we’re trying to archive the data, then they should be added to the netCDF file.

ggalibert:
Implemented by https://github.com/aodn/imos-toolbox/commit/8bff3ca09a0742fcad3b1b8a37f992567e81548b in v2.5.22

PR #398, opened 2016-11-24
Fix tooltip search string

sspagnol:
Not sure if this is an R2016b issue only, but the tool tip strings are 'Zoom In' and 'Pan', the ctrl short key text is never shown.


ggalibert:
Damn, my bad. I got confused and thought I could update the tooltip there while obviously it is retrieve the button...

Thanks for pointing out and sorry about that.

ggalibert:
Implemented by https://github.com/aodn/imos-toolbox/commit/3d85bbe883f08cebe9c0691eea52ca73d7632b5a in v2.5.22.

PR #399, opened 2016-11-25
legendflex modifcations to handle uipanel usage in figure creation

sspagnol:
Possible R2016b issue, but legendflex was failing to create new legend due to usage of axes being created in a uipanel object. Update legendflex.m to latest from https://github.com/kakearney/legendflex-pkg.git. Add modifications to handle the fact that figure are drawn in a uipanel by explicitly passing a 'parent' object reference (uipanel handle).


ggalibert:
It is rather annoying. The latest version of legendflex is not supported in my r2012b version.
The class inputParser doesn't have a addParameter attribute but instead a addParamValue. Will see if I can come up with a solution.

ggalibert:
Did you need latest version of legendflex to fix this or was explicit passing of a parent object enough?

ggalibert:
Ok, got it. Since r2013b, addParamValue is replaced by addParameter.

ggalibert:
Implemented by https://github.com/aodn/imos-toolbox/commit/b58e1179bea9365a550e25be7a72dfc80fce3210 in v2.5.22

PR #404, opened 2016-12-05
Commit Starmon DST sensor changes

petejan:
Changes to StarmonMini sensor to read the DST tilt sensor data files.

### DIFF
--- a/Parser/StarmonMiniParse.m
+++ b/Parser/StarmonMiniParse.m
@@ -16,6 +16,8 @@ function sample_data = StarmonMiniParse( filename, mode )
 %   sample_data - Struct containing sample data.
 %
 % Author: Guillaume Galibert <guillaume.galibert@utas.edu.au>
+% Contributor: Peter Jansen <peter.jansen@csiro.au> 
+%              Extend to read DST tilt sensor
 %
 
 %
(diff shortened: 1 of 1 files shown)


Parser/StarmonMiniParse.m
```diff
@@ -373,6 +386,7 @@
             data.(var).applied_offset = applied_offset;
             
         case 'Depth(m)'
+        case 'Depth0x28m0x29'
```
ggalibert: Do you mean:
```Matlab
case {'Depth(m)', 'Depth0x28m0x29'}
```
Otherwise the case 'Depth(m)' won't be handled anymore.

ggalibert:
Could you please attach a sample file?

Parser/StarmonMiniParse.m
petejan: Hmm, yes, matlab is not like C, where the the case drops through without a break

petejan:
Attached data file from DST tilt

[3X1010_saz15_star_oddi.zip](https://github.com/aodn/imos-toolbox/files/656124/3X1010_saz15_star_oddi.zip)

ggalibert:
Thanks, will be included in the next release.

ggalibert:
Can you please explain why Starmon DST can output in this format, or in the other format you suggested a new parser for (see https://github.com/petejan/imos-toolbox/commit/948f1c381d42ed9d01b3dfde9ae65cedf605d3cf)? Or are they two different instruments?

ggalibert:
Response from Pete:
> There are two types of instruments the DST Tilt and the DST CTD, the newer StarmonDSTparser should be able to read both formats, and even the older starmon mini-T format.

ggalibert:
Ok, then that would make sense to keep StarmonMiniParse.m as it was for Mini instruments (would assure compatibility with former Mini users) and introduce a new parser StarmonDSTParse.m for DST instruments, whether Tilt or just CTD (especially since the sample_data.meta.instrument_model value is hard coded in the parsers).

ggalibert:
Has been included in v2.5.23 vi commits https://github.com/aodn/imos-toolbox/commit/f9e692d333db7fed137b66cdf3b0529000156700 and https://github.com/aodn/imos-toolbox/commit/945881380faa219c652d4dd744b9de76e0228489

PR #417, opened 2017-02-15
create HG1/HG2 empty graphics handles array

sspagnol:


ggalibert:
Ok, can include this, thank you. In case you see this problem elsewhere please let me know.

I think when there is just 
```graphs = []```
or graphs is not even declared this doesn't cause a problem.

ggalibert:
Included in 2.5.25: https://github.com/aodn/imos-toolbox/commit/ac82886a7b498394ffe496f98d96da8e48b021d0

PR #418, opened 2017-02-15
Fix variable name inferred from list of ticked boxes

sspagnol:
UserData returned is in the form {line{k}, k} i.e. handle to a line object and an index of the plot, so previous code hTickBoxes(iTickBox) would only be the iTickBox index into all check boxes, not index into currently ticked boxes.


ggalibert:
Thanks for spotting this! Will include.

ggalibert:
Included in 2.5.25: https://github.com/aodn/imos-toolbox/commit/31ea1ca4debbb798ba02d25df0e9444de56b0050

PR #420, opened 2017-03-03
SBE39plus cnv firmware version and serial number

sspagnol:
Extract out FirmwareVersion and SerialNumber from cnv file produced by seabird software for SBE39plus instruments. Depending on model of the SBE39plus the cnv file doesn't have a xml like instrument header but is more like older style header with fields.

An example header looks like below
* Sea-Bird SBE39plus Data File:
* SerialNumber: 03907660
* FirmwareVersion: 4.2.0
* PressureInstalled: yes


ggalibert:
Sounds good, thank you. Will include in next release.

ggalibert:
Included in 2.5.25: https://github.com/aodn/imos-toolbox/commit/da2c67df68149d7f43e9d401fa94ca2b58611961

PR #434, opened 2017-06-01
Update flag resizing logic

sspagnol:
You could be really unlucky and have a short file (say due to instrument failure) and the number of ensembles equals the number of bins (of even number of spectral elements for wave measurement) so that original logic would produce an incorrect input for repmat.


ggalibert:
Thanks, will be included in next version of toolbox. The same logic that could be fixed is present in CTDSurfaceSoakQC.m (although less problematic since dealing with CTD casts).

PR #443, opened 2017-06-16
Handle Ruskin 1.13 date format change and update variable names.

sspagnol:
Newer version of Ruskin has change date format and added some new variable names.

For reference Ruskin 1.12 (Model=TGR-2050)
             Date & Time         Temp          Pres         Depth 
21-Nov-2016 04:59:40.000    26.1610417    10.0852656    -0.0468489 

Ruskin 1.13 (RBRduet)
             Date & Time       temp12        pres21        slop00        pres08        dpth01 
2016-11-21 05:00:00.000   27.3160538    10.1018553          null    -0.0306447    -0.0303948 

Added 'temp12' and 'pres21' as required.

Method to handle data format is simply to not specify a format and let Matlab datenum try. Tested on dd-mmm-yyyy and yyyy-mm-dd data sucessfully.

### DIFF
--- a/Parser/readXR620.m
+++ b/Parser/readXR620.m
@@ -247,10 +247,10 @@ function sample_data = readXR620( filename, mode )
                       data.(vars{k}) = data.(vars{k})/10;
                       
                       %Temperature (Celsius degree)
-                  case {'Temp', 'temp02'}, name = 'TEMP';
+                  case {'Temp', 'temp02', 'temp12'}, name = 'TEMP';
                       
                       %Pressure (dBar)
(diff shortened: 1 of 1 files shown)


ggalibert:
Sounds good, will be included in the next release. Thanks.

petejan:
I did find some time ago, that Ruskin had the nice trick of formatting the date based on the currently selected computer format, so can change based on what the user selects.

Need to be able to cope with any current sensible date formats.

ggalibert:
Hopefully with this version this will be the case since we leave Matlab's datenum to figure out what is the date format used.

PR #450, opened 2017-06-27
Fix unintialized struct

sspagnol:
psal was previous used as type real, but when returning from gsw_SP_from_R you are trying to create a struct.


ggalibert:
Thanks for that. Will be included in next release.

PR #451, opened 2017-07-06
Fix incorrect qcParam selection

sspagnol:
Previous logic iSample obtained from list of filenames (equivalent to number of elements in sample_data structure) and this is used to index into qcParam. Elements in qcParam are only calculated (and each element uniquely identified) by filename. But some parsers can return  more than one structure (eg awacParse). So eg load awac file numel(sample_data) ==2, but numel(qcParam) == 1. Change logic to just find the index into qcParam using passed in sample_data.toolbox_input_file

### DIFF
--- a/Graph/TimeSeries/graphTimeSeriesGeneric.m
+++ b/Graph/TimeSeries/graphTimeSeriesGeneric.m
@@ -66,9 +66,8 @@ set(ax, 'Tag', 'axis1D');
 
 % for global/regional range and in/out water display
 mWh = findobj('Tag', 'mainWindow');
-sMh = findobj('Tag', 'samplePopUpMenu');
-iSample = get(sMh, 'Value');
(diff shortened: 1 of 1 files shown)


ggalibert:
I cannot re-create such a problem with Awac velocity (.wpr) and wave data (.wap): numel(sample_data) ==2, and numel(qcParam) == 2 too. While I can see your code is valid and equivalent, I would to understand your motivations.

ggalibert:
What if you have two datasets from a single instrument file like it could be the case for Vector ADCPs?

If only looking at file names then first match is considered, not necessarily the right one...

ggalibert:
Ok, got it. The problem does occur with a Vector dataset. It is safe to gather all qcParam under same filename so then the problem raised above does not exist anymore. Will include you modifications in next release. Cheers.

PR #452, opened 2017-07-06
Change returning imosToolbox('version') logic.

sspagnol:
If user is only calling imosToolbox to get the version string don't bother setting paths etc.

### DIFF
--- a/imosToolbox.m
+++ b/imosToolbox.m
@@ -48,7 +48,14 @@ function imosToolbox(auto, varargin)
 % POSSIBILITY OF SUCH DAMAGE.
 %
 
+% Set current toolbox version
+toolboxVersion = ['2.5.28 - ' computer];
+
 if nargin == 0, auto = 'manual'; end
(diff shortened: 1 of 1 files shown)


ggalibert:
Sounds good. Will include this in next release. Thanks.

imosToolbox.m
```diff
@@ -48,7 +48,14 @@ function imosToolbox(auto, varargin)
 % POSSIBILITY OF SUCH DAMAGE.
 %
 
+% Set current toolbox version
+toolboxVersion = ['2.5.28 - ' computer];
+
 if nargin == 0, auto = 'manual'; end
+if nargin == 1 && strcmpi(auto, 'version')
```
ggalibert: Do you mind if I change this to simply:
```Matlab
if strcmpi(auto, 'version')
```
Since worst case 'auto' is being defined by the line just above.

sspagnol: Not at all. Just being paranoid :)

PR #453, opened 2017-07-06
Fix indexing into qcParam

sspagnol:
Similar logic to https://github.com/aodn/imos-toolbox/pull/451


ggalibert:
Will include this as well as similar logic for graphDepthProfileGeneric.m.

PR #463, opened 2017-08-07
Oxygen processing

ggalibert:


Parser/readWQMraw.m
```diff
@@ -87,9 +87,9 @@
   params{end+1} = {'temperature',     {'TEMP', ''}};
   params{end+1} = {'pressure',        {'PRES_REL', ''}};
   params{end+1} = {'salinity',        {'PSAL', ''}};
-  params{end+1} = {'DO(mg/l)',        {'DOX1_3', ''}};
-  params{end+1} = {'DO(mmol/m^3)',    {'DOX1_1', ''}};
-  params{end+1} = {'oxygen',          {'DOX1_2', ''}};
+  params{end+1} = {'DO(mg/l)',        {'DOX', ''}};
```
ggalibert: Should be DOXY if I understood your logic correctly.

Parser/readWQMraw.m
```diff
@@ -87,9 +87,9 @@
   params{end+1} = {'temperature',     {'TEMP', ''}};
   params{end+1} = {'pressure',        {'PRES_REL', ''}};
   params{end+1} = {'salinity',        {'PSAL', ''}};
-  params{end+1} = {'DO(mg/l)',        {'DOX1_3', ''}};
-  params{end+1} = {'DO(mmol/m^3)',    {'DOX1_1', ''}};
-  params{end+1} = {'oxygen',          {'DOX1_2', ''}};
+  params{end+1} = {'DO(mg/l)',        {'DOX', ''}};
+  params{end+1} = {'DO(mmol/m^3)',    {'DOXY', ''}};
```
ggalibert: Should be DOX1.

IMOS/imosParameters.txt
```diff
@@ -53,11 +53,13 @@ DIR_MAG,             0, from_direction,
...
+DOX2,                1, moles_of_oxygen_per_unit_mass_in_sea_water,                                               umol kg-1,     ,              ,                                  O, 999999.0, 0.0,      1000.0, float
 DOXS,                1, fractional_saturation_of_oxygen_in_sea_water,                                             percent,       ,              ,                                  O, 999999.0, ,         ,         float
 DOXY,                1, mass_concentration_of_oxygen_in_sea_water,                                                kg m-3,        ,              ,                                  O, 999999.0, 0.0,      29.0,     float
 DOXY_TEMP,           1, temperature_of_sensor_for_oxygen_in_sea_water,                                            degrees_Celsius,,             ,                                  T, 999999.0, 0.0,      50.0,     float
+OXSOL,               1, moles_of_oxygen_per_unit_mass_in_sea_water_at_saturation,                                 umol kg-1,     ,              ,                                  O, 999999.0, 0.0,      1000.0, float
```
ggalibert: "Is CF parameter" code should be 0, not 1. 
I could not find a CF standard name moles_of_oxygen_per_unit_mass_in_sea_water_at_saturation in http://cfconventions.org/Data/cf-standard-names/45/build/cf-standard-name-table.html . However, could this be a synonym of apparent_oxygen_utilization ("Apparent Oxygen Utilization (AOU) is the difference between measured dissolved oxygen concentration in water, and the equilibrium saturation concentration of dissolved oxygen in water with the same physical and chemical properties") ?

ggalibert: Maybe not AOU, since you seem to use gsw_O2sol_SP_pt() but then it seems to me it is the same thing as DOX2.

Preprocessing/oxygenPP.m
```diff
@@ -0,0 +1,248 @@
...
+  
+  psal = sam.variables{psalIdx}.data;
+  
+  ptmp = sw_ptmp(psal, temp , pres, 0);
+  
+  %SA = gsw_SA_from_SP(psal, pres, long, lat);
+  %pt = gsw_pt_from_t(SA, temp, pres, 0);
+  
+  % calculate the density at 0 dbar
+  dens = sw_dens0(psal, ptmp);
```
ggalibert: Not sure this is correct. Can you please explain:
1. Why density at 0bar?
2. If density at 0bar then why potential temperature?
3. If potential temperature then why not potential salinity?

Preprocessing/oxygenPP.m
```diff
@@ -0,0 +1,248 @@
...
+      coordinates = sam.variables{tempIdx}.coordinates;
+  else
+      coordinates = '';
+  end
+  temp = sam.variables{tempIdx}.data;
+  
+  % indexs of other variables that might be needed
+  psalIdx       = getVar(sam.variables, 'PSAL');
+  presRelIdx    = getVar(sam.variables, 'PRES_REL');
+  presIdx       = getVar(sam.variables, 'PRES');
```
ggalibert: Should be using this parameter when PRES_REL not available.

Preprocessing/oxygenPP.m
petejan: This is from http://archimer.ifremer.fr/doc/00287/39795/ (what argo is now doing).

> Unit conversion of oxygen
> The unit of DOXY is umol/kg in Argo data and the oxygen measurements are sent from Argo floats in another unit such as umol/L for the Optode and ml/L for the SBE-IDO. Thus the unit conversion is carried out by DACs as follows:
> O2 [umol/L] = 44.6596×O2 [ml/L]
> O2 [umol/kg] = O2 [umol/L] / ρ
> Here, ρ is the potential density of water [kg/L] at zero pressure and at the potential temperature (e.g., 1.0269 kg/L; e.g., UNESCO, 1983). The value of 44.6596 is derived from the molar volume of the oxygen gas, 22.3916 L/mole, at standard temperature and pressure (0°C, 1 atmosphere; e.g., García and Gordon, 1992). 
> This unit conversion follows the "Recommendations on the conversion between oxygen quantities for Bio-Argo floats and other autonomous sensor platforms" by SCOR Working Group 142 on "Quality Control Procedures for Oxygen and Other Biogeochemical Sensors on Floats and Gliders" [RD13] . 
> The unit conversion should always be done with the best available temperature, i.e., TEMP of the CTD unit.

The argument to sw_dens0 is practical salinity, is the practical salinity not conservative anyway.

IMOS/imosParameters.txt
petejan: OXSOL is the solubility of oxygen, not the measured value, your correct it is not a CF standard name.

Preprocessing/oxygenPP.m
petejan: So I need this code, from SalinityPP, can we carve this out into some function?
```
 % pressure information used for Salinity computation is from the
  % PRES or PRES_REL variables in priority
  if isPresVar
      if presRelIdx > 0
          presRel = sam.variables{presRelIdx}.data;
          presName = 'PRES_REL';
      else
          % update from a relative pressure like SeaBird computes
          % it in its processed files, substracting a constant value
          % 10.1325 dbar for nominal atmospheric pressure
          presRel = sam.variables{presIdx}.data - gsw_P0/10^4;
          presName = 'PRES substracting a constant value 10.1325 dbar for nominal atmospheric pressure';
      end
  else
      % when no pressure variable exists, we use depth information either 
      % from the DEPTH variable or from the instrument_nominal_depth 
      % global attribute
      if depthIdx > 0
          % with depth data
          depth = sam.(depthType){depthIdx}.data;
          presName = 'DEPTH';
      else
          % with nominal depth information
          depth = sam.instrument_nominal_depth*ones(size(temp));
          presName = 'instrument_nominal_depth';
      end
      
      % any depth values <= -5 are discarded (reminder, depth is
      % positive down), this allow use of gsw_p_from_z without error.
      depth(depth <= -5) = NaN;
      
      % pressure information needed for Salinity computation is either
      % retrieved from gsw_p_from_z when latitude is available or by 
      % simply assuming 1dbar ~= 1m
      if ~isempty(sam.geospatial_lat_min) && ~isempty(sam.geospatial_lat_max)
          % compute depth with Gibbs-SeaWater toolbox
          % relative_pressure ~= gsw_p_from_z(-depth, latitude)
          if sam.geospatial_lat_min == sam.geospatial_lat_max
              presRel = gsw_p_from_z(-depth, sam.geospatial_lat_min);
          else
              meanLat = sam.geospatial_lat_min + ...
                  (sam.geospatial_lat_max - sam.geospatial_lat_min)/2;
              presRel = gsw_p_from_z(-depth, meanLat);
          end
      else
          % without latitude information, we assume 1dbar ~= 1m
          presRel = depth;
          presName = [presName ' (assuming 1 m ~ 1 dbar)'];
      end
  end
  ```

Comment (ggalibert) [inline Preprocessing/oxygenPP.m]: Yep, will do.

File: IMOS/imosParameters.txt [hunk 6]
```diff
@@ -53,11 +53,13 @@ DIR_MAG,             0, from_direction,
...
+DOX2,                1, moles_of_oxygen_per_unit_mass_in_sea_water,                                               umol kg-1,     ,              ,                                  O, 999999.0, 0.0,      1000.0, float
 DOXS,                1, fractional_saturation_of_oxygen_in_sea_water,                                             percent,       ,              ,                                  O, 999999.0, ,         ,         float
 DOXY,                1, mass_concentration_of_oxygen_in_sea_water,                                                kg m-3,        ,              ,                                  O, 999999.0, 0.0,      29.0,     float
 DOXY_TEMP,           1, temperature_of_sensor_for_oxygen_in_sea_water,                                            degrees_Celsius,,             ,                                  T, 999999.0, 0.0,      50.0,     float
+OXSOL,               0, moles_of_oxygen_per_unit_mass_in_sea_water_at_saturation,                                 umol kg-1,     ,              ,                                  O, 999999.0, 0.0,      1000.0, float
```
ggalibert: Still not very happy with the suggested long_name. Why "at_saturation"?

If I understood well, by computing OXSOL based on potential temperature at atmospheric pressure then it is distinct from the DOX2 value at depth. So it would be a potential solubility at atmospheric pressure.

How about "moles_of_oxygen_per_unit_mass_in_sea_water_at_atmospheric_pressure" instead of the current long_name?

petejan: So this is what teos-10 function calls it,

```
USAGE:

O2sol = gsw_O2sol_SP_pt(SP,pt)

DESCRIPTION:

Calculates the oxygen, O2, concentration expected at equilibrium with  
air at an Absolute Pressure of 101325 Pa (sea pressure of 0 dbar) 
including saturated water vapor.  This function uses the solubility coefficients 
derived from the data of Benson and Krause (1984), as fitted by Garcia 
and Gordon (1992, 1993).

Note that this algorithm has not been approved by IOC and is not work 
from SCOR/IAPSO Working Group 127. It is included in the GSW
Oceanographic Toolbox as it seems to be oceanographic best practice.

INPUT:

SP  =  Practical Salinity  (PSS-78)                         [ unitless ]
pt  =  potential temperature                                   [ deg C ]
       with reference sea pressure (pr) = 0 dbar.

SP & pt need to have the same dimensions.

OUTPUT:

O2sol = solubility of oxygen                                 [ umol/kg ]
```

So could we just call it oxygen_solubility as a long name, with a comment calculated from teos-10 function gsw_O2sol_SP_pt(SP,pt)

ggalibert: It would be interesting to know what standard_name Argo is giving to this. In v3.2 of user manual, Argo is calling this parameter DOXY and gives the following attributes to it:
                DOXY:long_name = "Dissolved oxygen" ;
                DOXY:standard_name = "moles_of_oxygen_per_unit_mass_in_sea_water" ;
                DOXY:units = "micromole/kg" ;
                DOXY:valid_min = -5.f ;
                DOXY:valid_max = 600.f ;

Then IMOS already has a parameter code for that and it is DOX2. The only thing is that DOX2 was not computed properly before.

ggalibert: At the moment the toolbox cannot give both a long_name or a standard_name, it is one or the other.

petejan: Yes just a long name oxygen_solubility, no standard_name. I would almost not put it into the netCDF file as people can calculate it themselves should they want it from T & S. Its just needed to do a QC on the oxygen. To not put it into the netCDF file, do I just not put it into the imosParamters.txt?

ggalibert: Ok ok, now I get it, sorry about that I got pretty confused. 

Fine to have this param OXSOL with long_name "oxygen_solubility" and appropriate comment.

Preprocessing/oxygenPP.m
```diff
@@ -0,0 +1,200 @@
...
+  end
+  if (dox1Idx) && ~(dox2Idx)
+    dox1 = sam.variables{dox1Idx}.data;
+    dox2 = dox1 ./ dens;
+    dox2Comment = 'oxsolPP.m: derived from dox2 = dox (umol/l) / sw_dens0(PSAL, PTEMP)';
+  end
+  if (doxsIdx) && ~(dox2Idx)
+    doxs = sam.variables{doxsIdx}.data;
+    dox2 = doxs .* oxsol;
+    dox2Comment = 'oxsolPP.m: derived from doxs * oxsol(PSAL, TEMP)';
```
ggalibert: Could you please tell me where you got this formula from?

petejan: Saturation (from the sea bird manual for want of a better place) page 159 of 
http://www.seabird.com/document/sbe-data-processing-manual
```
Oxygen, percent saturation is the ratio of calculated oxygen to oxygen saturation, in percent:
(Oxygen / Oxygen saturation) * 100%.
```

Preprocessing/oxygenPP.m
```diff
@@ -0,0 +1,200 @@
...
+  end
```
ggalibert: What if a dataset contains all DOXS, DOX1 and DOX, does this mean there is an order of preference for computing DOX2?
1. from DOXS
2. from DOX1
3. from DOX

petejan: do we have reason be believe the manufacture calculates one better than the other? I would chose DOX (ml/l) from a SBE43 because that is what they are calibrated in. This would apply to a WQM data also. Optode they are calibrated in umol/l so DOX1 would be the preference.

ggalibert: Just asking because the current code implies there is an order of preference mentioned below. We can also imagine outputting DOX2_1, DOX2_2, etc.. depending on the data source and mention it in comment (I used to do that in the parsers I think).

Preprocessing/oxygenPP.m
ggalibert: Is that OK to do so when Seabird Oxsol formula Garcia and Gordon (1992) is not quite the same as the one used by GSW TEOS10 Benson and Krause (1984), as fitted by Garcia and Gordon (1992, 1993)?

petejan: Is this not the same in the underlying code? 
I would agree its preferable to have ml/l or umol/kg output from the instrument, and we should have the data processors provide that. Should the data provider only have % saturation, then they should know something about what they are doing before processing it into umol/kg.

test/test02/SBE16plus_01606330_2016_03_30.cnv
```diff
@@ -0,0 +1,42 @@
...
+# name 12 = v0: Voltage 0
+# name 13 = v1: Voltage 1
+# name 14 = v2: Voltage 2
+# name 15 = v3: Voltage 3
+# name 16 = v4: Voltage 4
+# name 17 = v5: Voltage 5
+# name 18 = potemp090C: Potential Temperature [ITS-90, deg C]
+# name 19 = sigma-�00: Density [sigma-theta, kg/m^3]
+# name 20 = flag:  0.000e+00
+
```
ggalibert: I'd be better if you could leave the original full header intact. At the moment too many information are missing for the file to be parsed properly.

Preprocessing/oxygenPP.m
```diff
@@ -0,0 +1,214 @@
...
+  
+  % check is psal and some dox measurement present, if not skip
+  if ~(psalIdx && (doxIdx || dox1Idx || dox2Idx || doxsIdx)), continue; end
+    
+  psal = sam.variables{psalIdx}.data;
+  
+  ptmp = sw_ptmp(psal, temp , presRel, 0);
+  
+  % calculate the density at 0 dbar
+  dens = sw_dens0(psal, ptmp);
```
ggalibert: I'm thinking about replacing calls to EOS-80 to equivalent in TEOS-10 in order to compute potential temperature and potential density. Something like:
```Matlab
  if ~isempty(sam.geospatial_lat_min) && ~isempty(sam.geospatial_lat_max) && ...
          ~isempty(sam.geospatial_lon_min) && ~isempty(sam.geospatial_lon_max)
      if sam.geospatial_lat_min == sam.geospatial_lat_max
          meanLat = sam.geospatial_lat_min;
      else
          meanLat = sam.geospatial_lat_min + ...
              (sam.geospatial_lat_max - sam.geospatial_lat_min)/2;
      end
      if sam.geospatial_lon_min == sam.geospatial_lon_max
          meanLon = sam.geospatial_lon_min;
      else
          meanLon = sam.geospatial_lon_min + ...
              (sam.geospatial_lon_max - sam.geospatial_lon_min)/2;
      end
  else
      disp(['Warning : no geospatial_lat_min/geospatial_lon_min documented for oxygenPP to be applied on ' sam.toolbox_input_file]);
      prompt = {'Latitude (South -ve)', 'Longitude (West -ve)'};
      dlg_title = 'Coords';
      num_lines = 1;
      defaultans = {'0', '0'};
      answer = inputdlg(prompt,dlg_title,num_lines,defaultans);
      
      % don't try to apply any correction if canceled by user
      if isempty(answer), continue; end
      
      meanLat = str2double(answer(1));
      meanLon = str2double(answer(2));
  end
  SA = gsw_SA_from_SP(psal, presRel, meanLon, meanLat);
  potTemp = gsw_pt0_from_t(SA, temp, presRel);
```
and
```Matlab
  CT = gsw_CT_from_pt(SA, potTemp);
  potDens = gsw_rho(SA, CT, 0);
```

petejan: Hmm, is the added complexity warranted?
The potential-temperature is a pre-teos-10 term, teos-10 developed a new term conservative temperature which is different to potential temperature. 
The argo document has,

> Here, ρ
> is the potential density of water [kg/L] at zero pressure and at the potential temperature (e.g.,  1.0269  kg/L;  e.g.,  UNESCO,  1983).  The  value  of  44.6596  is  derived  from  the  molar  volume  of  the oxygen  gas,  22.3916  L/mole,  at  standard  temperature  and  pressure  (0°C,  1  atmosphere;  e.g.,  García and Gordon, 1992). 

So I assumed its the EOS-80 defined potential density, do you have some indication that it should be calculated with teos-10?

test/test02/SBE16plus_01606330_2016_03_30.cnv
petejan: This is all that I thought that is needed to parse the file, this part of the file is the generated output from the sea-bird DataConversion utility, the rest of the header is from the instrument. 
I had not parsed the file with the toolbox in the test, but included it only for reference, are you planning on parsing this file?

Preprocessing/oxygenPP.m
ggalibert: Just thought that mixing EOS-80 and TEOS-10 wouldn't be that neat. 

TEOS-10 manual says:
> Potential   temperature   is   the   temperature   that   a   fluid   parcel  would  have  if  its  pressure  were  changed  to  a  fixed  reference  pressure   pr   in  an  isentropic  and isohaline   manner.
```Matlab
function pt0 = gsw_pt0_from_t(SA,t,p)

% gsw_pt0_from_t                               potential temperature with a
%                                       reference sea pressure of zero dbar
```

ggalibert: UNESCO 1983 has almost the same definition for potential temperature:
screenshot from 2017-08-16 15-05-53

Preprocessing/oxygenPP.m
```diff
@@ -0,0 +1,214 @@
...
+
+  % calculate oxygen solability
+  oxsol = gsw_O2sol_SP_pt(psal, ptmp); % sea bird calculates this using psal, temp, not potential temperature
```
petejan: How about we change this to the EOS-80 function then? Oh its in ml/l not Gordon and Garcia re-fit, I have see that code somewhere, oh its already under Util/O2sol.m maybe Ill change that then

ggalibert: Now that we process oxygen to consistent units via oxygenPP, was thinking about removing the code in rinkoPP that uses 02sol. Concentration would then be added by oxygenPP.

Preprocessing/oxygenPP.m
petejan: yes its a pre-teos-10 term.

Preprocessing/oxygenPP.m
petejan: I had not had a look at rinkoPP, is O2sol needed by the SBE43 processing also under Preprocessing\tranformation, hmm it has its own calculation. Looks like some more refactoring is needed :-)

ggalibert: This SBE43 processing is really old and I don't think any one is using it at IMOS so I'm happy to remove this.

test/test02/SBE16plus_01606330_2016_03_30.cnv
ggalibert: I had just tried out of curiosity and it's missing a description of flag bad. Something like:
```
# bad_flag = -9.990e-29
```
And also a description of the voltage columns would be welcome. Something like:
```
*             <Sensor id='volt 0'>
*                <Type>WHICHEVER DESCRIPTION</Type>
*                <SerialNum>NOT SET</SerialNum>
*                </Sensor>
```

Preprocessing/oxygenPP.m
petejan: Hmm so the O2sol.m function does not correct for ITS90 temperatures where as the gsw_O2sol_SP_pt does. Apart from that they both implement the Gordon and Garcia fit. I think I'll leave it as the teos-10 function.

ggalibert:
Necessary changes have been implemented in v2.5.32 of the toolbox via the following commits: 
https://github.com/aodn/imos-toolbox/commit/dc1a63b585a8657b26c3e76b98b95090c98c57ce
https://github.com/aodn/imos-toolbox/commit/9c3454b4184b53dc751bab423f9c6bc8aa7c4d08
https://github.com/aodn/imos-toolbox/commit/87aa02a181cc1e5b688573b7244de4a60a4be1d4
https://github.com/aodn/imos-toolbox/commit/06ffea2a6b90e51074229e892e35d006f4b9ad4b

PR #466, opened 2017-08-16
Handle alternate wap format

sspagnol:
Have had some processed wave data (wap file) that doesn't include all the extra statistics. Think it is due to instrument firmware version and if AST mode if available. The end result is the wap file has only 15 columns, but does contain all but one of the variables that is extracted from 31 column version (`SpectraType`).

While working on the other format, had Matlab errors on use of noninteger index from `nFreqFullSpectrum` calculation, and out of bounds messages on 
```matlab
waveData.fullSpectrum(i, :, :) = pwrFreqDir(start:start+nFreqFullSpectrum-1,:);
```
due to shorter wave file than expected.

### DIFF
--- a/Parser/readAWACWaveAscii.m
+++ b/Parser/readAWACWaveAscii.m
@@ -76,6 +76,23 @@ function waveData = readAWACWaveAscii( filename )
 %     30   Current direction (wave cell)    (degrees)
 %     31   Error Code
 %
+% OR%
+%      1   Month                            (1-12)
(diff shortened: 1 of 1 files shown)


ggalibert:
Looks good. Will include this in the next release. Cheers.

ggalibert:
Included in v2.5.32 via commit https://github.com/aodn/imos-toolbox/commit/ad46cbd9be38dcdc0d27d022a5afc18f27ff9b38

PR #469, opened 2017-08-22
Update list of required files for wave processing

sspagnol:
Update list of required files for wave processing and print out informative message if needed about which files are missing. The addition of whr file is for a future requirement this it might be the only way to detect if a user has applied a compass offset during wave processing.

The change of ```[path, name] = fileparts(filename);``` to ```[thePath, theName] = fileparts(filename);``` is a personal preference not to use variables names that are also function names.

### DIFF
--- a/Parser/readAWACWaveAscii.m
+++ b/Parser/readAWACWaveAscii.m
@@ -151,22 +151,30 @@ if ~ischar(filename), error('filename must be a string'); end
 waveData = [];
 
 % transform the filename into processed wave data filenames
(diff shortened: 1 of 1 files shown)


ggalibert:
Sounds good, will include in next release. Thanks!

ggalibert:
Included in v2.5.32 via commit https://github.com/aodn/imos-toolbox/commit/54d32a6a594d54ebc47bcbe7e0c4dea4d0c0b84a

PR #470, opened 2017-08-22
Capture directional/compass offset from whr file

sspagnol:
Dependent on PR [#469](https://github.com/aodn/imos-toolbox/pull/469) for import of whr file. Have followed simlar logic in workhorseParse.m in setting ```isMagBias```. Handling of parameters affected by this will have to be handled by the calling function.

### DIFF
--- a/Parser/readAWACWaveAscii.m
+++ b/Parser/readAWACWaveAscii.m
@@ -186,6 +186,25 @@ try
     pwrFreq    = importdata(pwrFreqFile);
     pwrFreqDir = importdata(pwrFreqDirFile);
     
+    % need to check if during waves processing has had compass/directional
+    % offset applied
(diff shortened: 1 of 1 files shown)


ggalibert:
Looks good thanks. The necessary update in awacParse to take into account this new information is missing but I will implement it. Will be included in the next release.

ggalibert:
Can we do the same for velocity files? At the moment we assume no magnetic corrreection has been applied. Even after processing with Storm / Quickwave Nortek software.

ggalibert:
Included in v2.5.32 via commit https://github.com/aodn/imos-toolbox/commit/a96fa0b3fc4e60deb436b68cf574091d576b1c58

sspagnol:
Sorry for the late reply been away. The problem is, for AWAC at least, if a correction has been applied it will only show up in the the wpb file (not the wpr file). We have just standardized on always loading the wpr file as is should have had no processing done to it (as it is a direct download from instrument).

PR #482, opened 2017-10-24
Close ResultSets and Statements

sspagnol:
Just noticed that JDBCDDB.java didn't close objects as per ODBCDDB.java


ggalibert:
Good idea, will include this in next release. Note that stmt and rs need to be declared next to conn at the beginning of the method so to be visible in the finally part of it.

PR #494, opened 2018-03-16
Add new 'prSM' pressure string identifier

sspagnol:
Was processing an old SBE19 SEACAT PROFILER V2.1a file and came across another pressure string, namely

# name 2 = prSM: Pressure, Strain Gauge [db]


ggalibert:
Thanks, will include in v2.5.37.

ggalibert:
Fixed via https://github.com/aodn/imos-toolbox/commit/1f4495e80cd07f91115f4c061070c9cc258e60ac

PR #504, opened 2018-04-30
Update readWQMraw.m to handle newer (v1.59) type header.

sspagnol:
The header produced downloading raw file from WQM with newer firmware (possibly also related to newer WQMHost) is different to what original code was written against (v1.26ish). I don't know exactly when the change to new type header occured. (1.2x to 1.5x?)

1. added some other serial number captures. Note it does seem as if these serial numbers aren't stored in final sample_data structure.
2. the original code assumed a default layout of data columns, but sometimes there is a `File Format:` string with which we can use to infer column indices. If there is no `File Format:` string the default ordering is assumed. Example strings are

```matlab
File Format: SN,State,Date,Time,Cond,Temp,Pres,RawDO,RawCHL,RawTurb,Volts
File Format: SN,State,Date,Time,Cond(S/m),Temp(C),Pres(dbar),RawDO(Hz),RawCHL(cts),RawTurb(cts),Volts
File Format: SN,State,Date,Time,Cond(S/m),Temp(C),Pres(dbar),RawDO(Hz),Pumped,CHLa(Counts),Turbidity(Counts),Volts
File Format: SN,State,Date,Time,Cond(S/m),Temp(C),Pres(dbar),RawDO(Hz),Pumped,CHLA(COUNTS),TURBIDITY(COUNTS),Volts
```
3. Updated comments on what `State` numbers I have seen. Have asked Seabird/Wetlabs on what they actually mean, but haven't had a response yet.
4. I have just done `try/catch` block of old/new layout. Could restructure to test for a few new layout strings and then put in `if/else` block, your choice.
5. Shortened example raw files attached [wqm_short_examples.zip](https://github.com/aodn/imos-toolbox/files/1959406/wqm_short_examples.zip)

### DIFF
--- a/Parser/readWQMraw.m
+++ b/Parser/readWQMraw.m
@@ -258,12 +258,16 @@ while ~strcmp(str{1},'<EOH>');
     headerLine(iline)=str{1};
 end
 
-wqmSNExpr    = ['^WQM SN: +' '(\S+)$'];
-ctdSNExpr    = ['^CTDO SN: +' '(\S+)$'];
-doSNExpr     = 'DOSN=(\S+)';
-opticsSNExpr = ['Optics SN: +' '(\S+)'];
+wqmSNExpr     = '^WQM SN:\s+(\S+)$';
+ctdSNExpr     = '^CTDO SN:\s+(\S+)$';
+doSNExpr      = '^DOSN=(\S+)';
+opticsSNExpr  = '^Optics SN:\s+(\S+)';
+doSNExpr2     = '^IDO SN:\s+(\S+)';
+opticsSNExpr2 = '^FL_NTU SN:\s+(\S+)';
+opticsSNExpr3 = '^ECO SN:\s+(\S+)';
+parSN         = '^PAR SN=(\S+)';
 
-exprs = {wqmSNExpr ctdSNExpr doSNExpr opticsSNExpr};
+exprs = {wqmSNExpr ctdSNExpr doSNExpr opticsSNExpr doSNExpr2 opticsSNExpr2 opticsSNExpr3 parSN};
 
 for l = 1:length(headerLine)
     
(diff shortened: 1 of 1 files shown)


sspagnol:
Meant to also say, I don't know know what a new style header with PAR looks like, I only have seen older v1.26 style.

sspagnol:
Had a response from Seabird/Wetlabs

RID - Record ID Definitions 

Record ID | Function | Comments
-|-|-
4 | Raw Meter Data during DO Stabilization | Not used in .DAT files
5 | Raw Meter Data during Optical Startup | Not used in .DAT files
6 | Raw Meter Data during Sample | Used in .DAT files
100 |Status Record | 
110 |UPS Activated V1 |Obsolete in v1.5x
111 |UPS Activated V2 |Obsolete in v1.5x
112 |UPS Activated V3 |
120 |Invalid Command Received |
130 |Error Message in Text Format |
140 |Generic Text Message |
200 |Raw CTD Sensor Record |
201 |Invalid Raw CTD Sensor Record |
210 |Raw ECO Sensor Record |
211 |Invalid Raw ECO Sensor Record |
220 |Raw EDP Sensor Record |
221 |Invalid Raw EDP Sensor Record |

And as had already been worked out, RID: 4, 5, and 6 WQM Raw Data Record
Example:
147,6,032011,064449,5.49246,26.5499,16.27,10315.2,1,101,95,16.10

Text | Value | Comment
-----|-------|-----------
 WQM SN | 147 |
 Record ID | 6 | Operational State
 Date | 032011 | MMDDYY
 Time | 064449 | HHMMSS
 Conductivity | 5.49246 | S/m
 Temperature | 26.5499 | ITS-90
 Pressure | 16.27 | dbar
 Raw Dissolve Oxygen | 10315.2 | Hz
 Pump Status | 1 | 0=off, 1=slow, 2=fast
 Raw ECO Signal 1 | 101 | Counts
 Raw ECO Signal 2 | 95 | Counts
 Voltage: | 16.10 | Volts

The document has further info on each RID but it is still in preliminary status so isn't released. When we get a final copy will attach it to the wiki.

petejan:
Where is the updated test case? with example data files included?

ggalibert:
I will put a test case together for this. 

@sspagnol were all the example files v1.59 useful? I can see that only 2 distinct file formats are represented:
`SN,State,Date,Time,Cond(S/m),Temp(C),Pres(dbar),RawDO(Hz),Pumped,CHLA(COUNTS),TURBIDITY(COUNTS),Volts`
and
`SN,State,Date,Time,Cond(S/m),Temp(C),Pres(dbar),RawDO(Hz),Pumped,CHLa(Counts),Turbidity(Counts),Volts`

Also, did you realise `WQM0064_003.Raw` and `WQM0067_002.Raw` had 2 sets of header/data? Is the parser working for them like you expect it to?

PR #523, opened 2018-11-01
Set axis in colorbar contextmenu callback.

sspagnol:
In at least R2017a onwards the use of gca in setTimeSerieColorbarContextMenu callback (cbCLimRange) results in wrong plot color range being changed. Also not specifiying an axis also resulted in wrong colormap being applied (or being applied to all plots). 

The use of 
```
if verLessThan('matlab', '8.4')
    cb = colorbar('peer', ax);
else
    cb = colorbar(ax);
end
```

is just a recommendation in R2018b manual, it doesn't yet issue a warning yet.

### DIFF
--- a/Graph/TimeSeries/graphTimeSeriesTimeDepth.m
+++ b/Graph/TimeSeries/graphTimeSeriesTimeDepth.m
@@ -76,14 +76,19 @@ posWithoutCb = get(ax, 'Position');
 
 h = pcolor(ax, double(xPcolor), double(yPcolor), double(var.data'));
 set(h, 'FaceColor', 'flat', 'EdgeColor', 'none');
-cb = colorbar('peer', ax);
+% 'peer' input is not recommended starting in R2014b
+if verLessThan('matlab', '8.4')
+    cb = colorbar('peer', ax);
+else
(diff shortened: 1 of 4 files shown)


ggalibert:
Thanks, I will include this in 3 separate commits and see if there's a need to extend those changes to other scripts.

Graph/TimeSeries/graphTimeSeriesTimeDepth.m
```diff
@@ -96,7 +101,7 @@
         'Layer',        'top', ...
         'Tag',          'axis2D');
 
-if all(all(colormap == rkbwr))
```
ggalibert: Can you please explain this change?

ggalibert: Ah ok, I think I got it. Same pb as in setTimeSerieColorbarContextMenu.m for version >= r2017a ?

ggalibert:
> And is there a reason why the routine is called "setTimeSerieColorbarContextMenu" rather than "setTimeSeriesColorbarContextMenu" (i.e TimeSerie vs TimeSeries)

It is a typo. Can fix this.

PR #524, opened 2018-11-23
Fix incorrect structure assignment in wave structure

sspagnol:
fix incorrect structure access


ggalibert:
Thanks, will include in next release.

PR #555, opened 2019-06-28
Handle case of WQM which didn't record one of it's primary variables

sspagnol:
Had a case of a very badly behaved WQM which didn't record one of it's primary variables (press in my case). So just a simple check that cond/pres/temperature aren't empty. But it does have flow on effects for calculation of salinity and oxygen.

Wondering if same test should also be done for oxygen and O2cal depend on ~noSainity || ~noOxygen?

Have coded missing variable logic as (eg) noTemperature is true if no temperature, but you may wish to rewrite as hasTemperature is true if have temperature.

### DIFF
--- a/Parser/readWQMraw.m
+++ b/Parser/readWQMraw.m
@@ -567,18 +567,50 @@ end
 % reshape it to make it easier to extract data
 A = A';
 
-% C,T,P are in stored in engineering units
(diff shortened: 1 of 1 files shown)


sspagnol:
And maybe disply message on console that this had happened.

PR #561, opened 2019-07-18
Sspagnol wqm missingvar, closes #557

### LINKED ISSUE #557: Handle case of WQM which didn't record one of it's primary variables
See PR https://github.com/aodn/imos-toolbox/pull/555 .

@sspagnol can you please upload a zip file of an example of this misbehaved instrument file?

ocehugo:
These are the fixes done on top of PR #555. I had to implement some extras.

I had to refactor since the scope was a bit messy. I think I left the code in better shape.

The PR can generate the structures for all files within the test folder and runTests.m got a pass.

### DIFF
--- a/Parser/readWQMraw.m
+++ b/Parser/readWQMraw.m
@@ -1,16 +1,16 @@
 function sample_data = readWQMraw( filename, mode )
 %readWQMraw parses a .RAW file retrieved from a Wetlabs WQM instrument.
 %
-% This function is able to parse raw data retrieved from a Wetlabs WQM CTD/ECO 
-% instrument. 
+% This function is able to parse raw data retrieved from a Wetlabs WQM CTD/ECO
+% instrument.
 %
 % Inputs:
 %   filename    - name of the input file to be parsed
 %   mode        - Toolbox data type mode.
 %
 % Outputs:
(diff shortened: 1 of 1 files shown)


ocehugo:
@sspagnol

ocehugo:
argh, those binary files + build stuff were not supposed to be there. I mixed match when changing branches.

ocehugo:
@ggalibert ,
cleaned up

ocehugo:
PS: I did a squash of three commits because that is the right thing to do - you want a clean log on the repo, not a log with every hunk committed.

Unfortunately, the diffs are big because the file is messy. windows formatting, leading white spaces, no consistent indentation, large scope. I use auto-lint so a lot of noise was generated. There still some indentation problems though

ggalibert:
> PS: I did a squash of three commits because that is the right thing to do - you want a clean log on the repo, not a log with every hunk committed.

Ok but for reviewing it is a lot harder!!! You could squash after review and before merging.

ggalibert:
For discussion: In the past I used to only merge to master when I was doing a release, so merging a whole bunch of commits + updated binaries at once. The reason was some users would checkout master at any time of dev and use the code to process their data instead of the 2.5 official branch. The actual version of the toolbox used would then be lost.

PR #572, opened 2019-07-30
support flCUVA+PAR in SBE19Plus cnv, closes #550

### LINKED ISSUE #550: New SeaBird code for Chlorophyl-a
In the .cnv file, a new code header for chlorophyl-a `flCUVA` has been introduced and is currently not supported. `convertSBEcnvVar` needs to be updated accordingly.

ocehugo:
Please see discussion in #550. 

This should require a proper test but the codebase is not ready yet for this

### DIFF
--- a/Parser/convertSBEcnvVar.m
+++ b/Parser/convertSBEcnvVar.m
@@ -122,12 +122,23 @@ switch name
     case 'flC'
       name = 'CPHL';
       comment = 'Artificial chlorophyll data computed from bio-optical sensor raw counts measurements. Originally expressed in ug/l, 1l = 0.001m3 was assumed.';
-      
+
+    %fluorescence (ug/l)
+    %Aquatracka SBE19plus cdom sensor as cphl implemented on #550
+    case 'flCUVA'
+      name = 'CPHL';
+      comment = 'Artificial chlorophyll data computed from bio-optical sensor raw counts measurements (Chelsea UV Aquatracka). Originally expressed in ug/l, 1l = 0.001m3 was assumed.';
+
+    % PAR/Irradiance, Biospherical/Licor
+    %par/sat/log - light sensor in SBE19plus assumed in umol photons/m2/s and implemented on #550
+    case {'par0x2Fsat0x2Flog', 'par/sat/log'}
+      name = 'PAR';
+      comment = 'PAR/Logarithmic/ Satlantic';
+
     % artificial chlorophyll from fluorescence (mg/m3)
     case 'flECO0x2DAFL'
       name = 'CPHL';
       comment = 'Artificial chlorophyll data computed from bio-optical sensor raw counts measurements.';
-      
     % oxygen (mg/l)
     % mg/l
     case 'sbeox0Mg0x2FL'


Parser/convertSBEcnvVar.m
```diff
@@ -122,12 +122,21 @@
     case 'flC'
       name = 'CPHL';
       comment = 'Artificial chlorophyll data computed from bio-optical sensor raw counts measurements. Originally expressed in ug/l, 1l = 0.001m3 was assumed.';
-      
+
+    %Aquatracka SBE19plus new chla #550
+    case 'flCUVA'
+      name = 'CPHL';
+      comment = 'Chlorophyll estimated from fluorescence. No coeficients provided.  Chelsea UV Aquatracka assumed in ug/l, from SBE19Plus.';
```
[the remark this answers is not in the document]
ocehugo: I changed the comments for two reasons:

1. convertSBEcnvVar is used by both SBE37 and SBE19. This is why I put SBE19Plus in the comment - to differentiate in the case where SBE37 has flCUVA. If this is not possible in any case, then I should remove; otherwise, the removal can create provenance tracking problems.

2. There is no way to differentiate "flC" of "flCUVA". They probably are measured by different sensors. I got no clue of what the respective sensor for "flC" from the code. I would like to know when concatenating CPHL that they came from different sensors and who is who. The comments are not the optimal way, but ATM is the only way to do that.

If I copy the comment, we lost provenance of "Aquatracka" sensor.  

I can add the units as comments ( which should be read but this is another story). However,  I want to keep the reference of the issue since this is extra functionality (and is easy to grep).

Parser/convertSBEcnvVar.m
```diff
@@ -122,12 +122,23 @@
     case 'flC'
       name = 'CPHL';
       comment = 'Artificial chlorophyll data computed from bio-optical sensor raw counts measurements. Originally expressed in ug/l, 1l = 0.001m3 was assumed.';
-      
+
+    %fluorescence (ug/l)
+    %Aquatracka SBE19plus cdom sensor as chpl #550
+    case 'flCUVA'
+      name = 'CPHL';
+      comment = 'Artificial chlorophyll data computed from bio-optical sensor raw counts measurements (Chelsea UV Aquatracka mounted on SBE19Plus). Originally expressed in ug/l, 1l = 0.001m3 was assumed.';
```
ggalibert: You don't need to include provenance from SBE19Plus. This is captured in global attribute + flCUVA code is going to be used whether the sensor is mounted on an SBE19 or SBE35.

Parser/convertSBEcnvVar.m
```diff
@@ -122,12 +122,23 @@
...
+
+    % PAR/Irradiance, Biospherical/Licor
+    %PAR/sat/log - light sensor in SBE19plus  #550
+    case {'par0x2Fsat0x2Flog', 'par/sat/log'}
+      name = 'PAR';
+      comment = 'PAR/Logarithmic/ Satlantic, assumed in umol photons/m2/s, from SBE19Plus.';
```
ggalibert: Same here, can remove mention to SBE19Plus.

ocehugo:
amended

Parser/convertSBEcnvVar.m
ocehugo: amended

ocehugo:
amended

Parser/convertSBEcnvVar.m
```diff
@@ -122,12 +122,23 @@
...
+    %Aquatracka SBE19plus cdom sensor as chpl implemented on #550
+    case 'flCUVA'
+      name = 'CPHL';
+      comment = 'Artificial chlorophyll data computed from bio-optical sensor raw counts measurements (Chelsea UV Aquatracka). Originally expressed in ug/l, 1l = 0.001m3 was assumed.';
+
+    % PAR/Irradiance, Biospherical/Licor
+    %par/sat/log - light sensor in SBE19plus implemented on #550
+    case {'par0x2Fsat0x2Flog', 'par/sat/log'}
+      name = 'PAR';
+      comment = 'PAR/Logarithmic/ Satlantic, assumed in umol photons/m2/s';
```
ggalibert: Please remove mention of unit there. You can put this information in commented code if you want.

Parser/convertSBEcnvVar.m
```diff
@@ -122,12 +122,23 @@
     case 'flC'
       name = 'CPHL';
       comment = 'Artificial chlorophyll data computed from bio-optical sensor raw counts measurements. Originally expressed in ug/l, 1l = 0.001m3 was assumed.';
-      
+
+    %fluorescence (ug/l)
+    %Aquatracka SBE19plus cdom sensor as chpl implemented on #550
```
ggalibert: cphl

Parser/convertSBEcnvVar.m
ocehugo: amended again

PR #573, opened 2019-08-08
add ignorelist + enable arguments on compile call

ocehugo:


Util/imosCompile.m
```diff
@@ -44,9 +44,17 @@ function imosCompile()
...
-exportRoot    = pwd;
-toolboxRoot   = [exportRoot filesep '..' filesep '..'];
-stagingRoot   = [exportRoot filesep '..' filesep 'staging'];
+if nargin < 1,
+  exportRoot    = pwd;
+  toolboxRoot   = [exportRoot filesep '..' filesep '..'];
+  stagingRoot   = [exportRoot filesep '..' filesep 'staging'];
+elseif nargin == 1,
+  exportRoot = toolboxRoot;
+  stagingRoot = [ toolboxRoot '/' 'dist'];
```
ggalibert: Use `filesep` instead of '/' so that it is platform independent.

Another way to do it is `stagingRoot = fullfile(toolboxRoot, 'dist');`

ocehugo: amended

Issue #568, opened 2019-07-22
Master merging rule


ocehugo:
From @ggalibert :

>For discussion: In the past I used to only merge to master when I was doing a release, so merging a whole bunch of commits + updated binaries at once. The reason was some users would checkout master at any time of dev and use the code to process their data instead of the 2.5 official branch. The actual version of the toolbox used would then be lost.

The above is a problem because the version is hard-coded. Git tags should set the versions. Hence, the toolbox build code needs to interact with git. 

Also, master is master, i.e. is the place of dev/latest things.  If people like to live on the edge, they run master. If they want a stable release, they get the tag release.

The toolbox version should always be present - even if running from master. In this case, the version should be appended with the latest commit if the code is clean (i.e. is the exact commit). If the code is dirty (not staged stuff), they should receive a "dirty" string. This is the standard in python packaging, for example.

In particular, the new alternative build script at #564 write a build log to ```log_v2.5.41-16-gaa14381```.  If there is dirty stuff, it will get ```log_v2.5.41-16-dirty-gaa14381```. 

This way, we can also track files built with custom code, which should be discouraged.

ggalibert:
> This way, we can also track files built with custom code, which should be discouraged.

It would be great if we could have the pipeline/checker detecting the use of a non staged version in order to  not allow data upload with non staged versions of the toolbox. Liaise with @mhidas .

petejan:
Hmm, this is a bad idea, often have to hack the toolbox to process the diverse files I use. 
We also generate many files without using the toolbox, for example realtime files, gridded product files.

mhidas:
> It would be great if we could have the pipeline/checker detecting the use of a non staged version

This could be done as part of an ANMN-specific check suite (yet to be created), which would not be applied to @petejan's files.

sspagnol:
Question (keeping in mind I'm not a git expert) : currently we (AIMS) have imos-toolbox repository with master (follows AODN master), 2.5 (follows AODN 2.5 branch) and AIMS-2.5 branches. The AIMS-2.5 branch has AIMS mods (other parsers, settings etc). When a new 2.5.x release happens, I pull AODN 2.5 into local 2.5 branch (so I check out what has changed). Then merge 2.5 into AIMS-2.5. 

So what would be the state of the tag release now?

ocehugo:
[...]

PR #576, opened 2019-08-20
[WIP] include new python build script

ocehugo:
This is a new build script for the stand-alone version for discussion.

Advantages:
- No matlab calls, except mcc.
- better capabilities: specify the MCCRoot, repo folder, binary output folder and the architecture.
- improve binary names to include current repository state (good for debugging, verification, sharing binaries, etc).
- write auxiliar version number to .standalone_canonical_version (standalone requirements)

The script includes the same logic/capability mentioned in #568, but it's not the entire solution. Some logic is still required in the toolbox for stand-alone and the non-binary usages. However, the idea is here - replicate the get_info method - and set the version code accordingly - this is discussed further in #568.

To discuss:
- Should we keep maintaining the binary names without version number?
- I like to be pedantic in regard to the architecture but this can be dropped/adapted for automatic detection if required.

### DIFF
--- a/build.py
+++ b/build.py
@@ -1,0 +1,221 @@
+#!/usr/bin/env python3
+""" Build the imos stand-alone toolbox GUI using only the Matlab compiler.
+
+This script ignores untracked and testfiles within the repository and weakly
+attach a string version to the binary filename.
+The string is related to the repo state.
+
+Usage:
+  build.py  --arch=<architecture> [--root_path=<imostoolboxpath> --mcc_path=<mccpath> --dist_path=<distpath>]
+
+Options:
+  -h --help     Show this screen.
+  --version     Show version.
+  --mcc_path=<mccpath>           The Matlab runtime Path
+  --root_path=<imostoolboxpath>  The repository root path
+  --dist_path=<distpath>  Where Compiled items should be stored
+  --arch=<architecture>   One of win64,glnxa64,maci64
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -0,0 +1,221 @@
...
+Usage:
+  build.py  --arch=<architecture> [--root_path=<imostoolboxpath> --mcc_path=<mccpath> --dist_path=<distpath>]
+
+Options:
+  -h --help     Show this screen.
+  --version     Show version.
+  --mcc_path=<mccpath>           The Matlab runtime Path
+  --root_path=<imostoolboxpath>  The repository root path
+  --dist_path=<distpath>  Where Compiled items should be stored
+  --arch=<architecture>   One of win32,glnxa64,maci64
```
ggalibert: win32 is not possible I believe, instead win64.

build.py
```diff
@@ -0,0 +1,221 @@
...
+
+from docopt import docopt
+from git import Repo
+
+VALID_ARCHS = ['glnxa64', 'win64', 'maci64', 'win32']
+ARCH_NAMES = {
+    'glnxa64': 'Linux64',
+    'maci64': 'Mac64',
+    'win64': 'Win64',
+    'win32': 'Win32'
```
ggalibert: win32 is not available in R2018 anymore.

PR #578, opened 2019-08-21
schema validation for most matlab types, closes #577

### LINKED ISSUE #577: Schema validation for matlab variables
It's pretty easy to know if a variable is different from another in Matlab. However, it's not so straight forward to see the reason.

This is a problem when dealing with new Parsers, for example. One seeks to check if the new parser produces the same structure after a modification. The canonical way is to just:

```isequal(new,old)```

However, this does not tell where and why the difference occurs. Also, it's useful not just to know where the content changes, but why. Hence, a useful tool is something like:

``` [isdiff,why] = compare_this(new,old)```

Type, size, length, and content comparison is required to achieve that and inform correctly. For example, a given structure got an extra field not previously found (length/size wise comparison), the type at new.data.cell{3} can be ```char``` while in old.new.data{3} can be ```int16```, etc.

Finally, the comparison needs also to walkthrough all nested levels, reporting exactly where/why the difference occurs (i.e. recursion).

The above requirements indicate that what we need is a schema validation tool.

Requirements:
- [x] A schema validation toolbox that support both flat (str,array) and nested/tree-like variables (struct/cell).
- [x] The tool should be recursive and extensible, so both schemas creation and schema validation can be performed.
- [x] The function has to report correctly why and where the differences are found, in arbitrary and deeply nested trees (e.g. ```a.b{3}.c{3,3}.d(4,1,3).e{3}.f```).
- [x] Be extensible, so we can build schemas and/or evaluate different aspects.

### PR DESCRIPTION
**Pull request by ocehugo**
This PR includes a schema validation utility.

It's included here as the primary tool for me (and hopefully others) to develop the toolbox with further confidence and provide enhanced checking when comparing and ensuring the data structures are what they should be.

This PR is self-contained and do not depend on anything related to the toolbox - only Matlab standard routines.

### DIFF
--- a/Util/Schema/createTree.m
+++ b/Util/Schema/createTree.m
@@ -59,7 +59,7 @@ function [outcome, root, nrec] = createTree(arg1, func, outcome, name, nrec),
     % along with this program.
     % If not, see <https://www.gnu.org/licenses/gpl-3.0.en.html>.
     %
-    
+
     narginchk(1, 5)
 
     if nargin < 2,
--- a/Util/Schema/createTree.m
+++ b/Util/Schema/createTree.m
@@ -1,0 +1,74 @@
+function [stype, root, nrec] = createTree(arg1, stype, name, nrec, func),
+    % function [stype, root, nrec] = createTree(arg1, stype, name, nrec, func),
+    % A function to recursively apply a function handler @func,
+    % to all fields of an argument.
+    % the function generate the exact same type of arg1,
+    % visiting/applying @func through all nested structures (cells/structs)
+    %
+    % `arg1` = a matlab variable
+    % stype = the parent structure - internal use
+    % name = the name of this walk level - internal use
+    % nrec = the level of this walk level - internal use
+    % func = a special function handle that return 3 outputs - see DetectType
+    % <->
+    % stype = the result structure with follow-up nests/walks.
+    % root = is the result of @func for the root level of arg1.
+    % nrec = is the number of children visits.
+    %
+    % author: hugo.oliveira@utas.edu.au
+    if nargin < 2,
+        stype = struct();
+        name = 'root';
+        nrec = -1;
+        func = @detectType;
+    elseif nargin < 3,
+        name = 'root';
+        nrec = -1;
+        func = @detectType;
+    elseif nargin < 4,
+        nrec = -1;
+        func = @detectType;
+    elseif nargin < 5,
+        func = @detectType;
+    end
+
+    stack = dbstack;
+    selfname = stack.name;
+    self = str2func(selfname);
+
+    iname = name;
+
+    [stype, is_nested, nested_keys] = func(arg1);
+    first_level = stype;
+
+    if is_nested,
+
+        if isstruct(stype),
+            fnames = fields(stype);
+            nested_names = fnames(nested_keys);
+
+            for k = 1:length(nested_names),
+
+                kname = nested_names{k};
+                data = arg1.(kname);
+                [stype.(kname), root, nrec] = self(data, stype, kname, nrec - 1, func);
+            end
+
+        elseif iscell(stype),
+
+            for k = 1:length(nested_keys),
+                data = arg1{k};
+                [stype{k}, root, nrec] = self(data, stype, iname, nrec - 1, func);
+            end
+
+        end
+
+    end
+
+    %adjust nrec and root
+    if strcmpi(iname, 'root'),
+        nrec = abs(nrec) - 1;
+    end
+    root = first_level;
+
+end
--- a/Util/Schema/createTree.m
+++ b/Util/Schema/createTree.m
@@ -6,14 +6,14 @@ function [stype, root, nrec] = createTree(arg1, stype, name, nrec, func),
     % visiting/applying @func through all nested structures (cells/structs)
     %
     % `arg1` = a matlab variable
-    % stype = the parent structure - internal use
-    % name = the name of this walk level - internal use
-    % nrec = the level of this walk level - internal use
-    % func = a special function handle that return 3 outputs - see DetectType
+    % `stype` = the parent structure - internal use
+    % `name` = the name of this walk level - internal use
+    % `nrec` = the level of this walk level - internal use
+    % `func` = a special function handle that return 3 outputs - see DetectType
     % <->
-    % stype = the result structure with follow-up nests/walks.
-    % root = is the result of @func for the root level of arg1.
-    % nrec = is the number of children visits.
+    % `stype` = the result structure with follow-up nests/walks.
+    % `root` = is the result of @func for the root level of arg1.
+    % `nrec` = is the number of children visits.
     %
     % author: hugo.oliveira@utas.edu.au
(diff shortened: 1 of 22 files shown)

### CODE REVIEW

File: Util/Schema/createTree.m [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+    % stype = the parent structure - internal use
+    % name = the name of this walk level - internal use
+    % nrec = the level of this walk level - internal use
+    % func = a special function handle that return 3 outputs - see DetectType
+    % <->
+    % stype = the result structure with follow-up nests/walks.
+    % root = is the result of @func for the root level of arg1.
+    % nrec = is the number of children visits.
+    %
+    % author: hugo.oliveira@utas.edu.au
```
ggalibert: Can you please re-use the header template consistently across new scripts? This includes licencing info. Also, try to adopt the same indent style as with historical functions if possible.

Util/Schema/detectCellContent.m
```diff
@@ -0,0 +1,29 @@
+function [itype, is_nested, nested] = detectCellContent(arg1);
```
ggalibert: Needs a header

Util/Schema/detectStructContent.m
```diff
@@ -0,0 +1,37 @@
+function [itype, is_nested, nested] = detectStructContent(arg1);
```
ggalibert: Needs a header

Util/Schema/isdouble.m
```diff
@@ -0,0 +1,3 @@
+function [bool] = isdouble(arg),
```
ggalibert: Do we really need all these functions (11 already!)? A bit overkill when isa() already is a one liner. I'm just a bit concerned that we'll end up with crazy amount of new .m files in the repo if we follow this practice consistently.

Util/Schema/whichtype.m
```diff
@@ -0,0 +1,11 @@
+function [argtype] = whichtype(arg),
```
ggalibert: Do we need this?

Util/Schema/createTree.m
ocehugo: @ggalibert , I was supposed to say something about that but forgot. I don't think the license should be in every single m file. The license should be on the repository - all files within the repo should abide by the same license. This adds too much inflation to the codebase. I'm happy to add but it's just something I found useless

Util/Schema/isdouble.m
ocehugo: It's not my fault that matlab requires functions to be in separated files if you want to use then outside of the current function scope. The main reason, however, is that I need a static function handle for type definition (see detectType). 

The only way to do that is to define a proper function, so the function handle to it is static and comparable. Moreover, these other types do not have proper signatures such as ischar.

Example:

In detectType(A) we would define:
```
isdouble = @(x) isa(x,'double')
...
Atype = @isdouble
```

In detectType(B) we would do:
```
isdouble = @(x) isa(x,'double')
...
Btype = @isdouble
```

Comparison scope would fail, because dynamical function handles are assigned to different addresses in memory:
```
Atype = detectType(A)
BType = detectType(B)
isequal(Atype,Btype) == false
```

When, I define a proper/static function, the above works in any scope:

```
 Atype = @isdouble
 Btype = @isdouble
 isequal(Atype,Btype) == true
```

ocehugo:
>-I assumed you checked performance of this. I note that there are a few for loops and this can be problematic(some of them might be vectorised).

No, my calls and tests are fast enough that I didn't even bother looking into it. Calls with very deep nested structures usually take less than 10ms.  The entire 102 tests take 7s in my machine (<10ms per test), accounting for the overhead of the test suite (setup, defs, loads,etc).

Vectorization is not very applicable - I'm not dealing with fixed-sized arrays but with unknown nested levels of structs/cells in recursive mode (tree walk with unknown deep levels). Modern matlab also got a good jit compiler and for loops are sometimes faster than vectorized versions.

Util/Schema/detectCellContent.m
ocehugo: done

Util/Schema/detectStructContent.m
ocehugo: done

Util/Schema/whichtype.m
ocehugo: Yes - It's an good alias and to define a string instead of getting the 5th argument of objinfo. It what you would use to get a string representation of a type.

Util/Schema/isdouble.m
ggalibert: Could you declare your function isdouble etc in detectType instead of in separate .m files for each of them?

Util/Schema/whichtype.m
ggalibert: I mean, again it's a one liner in a separate .m file...

ocehugo:
I just finished adding some missing help strings.

Util/Schema/isdouble.m
ocehugo: In general: No. One may need them in any scope, not just detectType. The function handle of that function is also an output. Hence, it should easily reachable.

Specifically, they are used in testSchema. For example for input and expected output when testing. They cannot be dynamic function handles otherwise the comparison would fail.

Util/Schema/whichtype.m
ocehugo: That's a feature!

Would you prefer reading:

```[~,~,~,~,type] = objinfo(a)```

instead of 

```type = whichtype(a)```

?

Note also that objinfo uses objinfo (one level recursion). You didn't notice because ```whichtype``` clarify the code intent, otherwise we would have  ```[~,~,~,~,otype{k}] = objinfo(data)``` calls instead.

To delete ```whichtype``` is also to sacrifice expressiveness for code/scope size. For example, everytime you want a type you have to inspect objinfo help to remember it's the 5th output argument, instead of just using a simple function that actually output only what you want.

Being a oneliner is a feature: I'm reusing the code in objinfo to create a function that is a simplified version not available anywhere, with a proper name that matches intent.

ocehugo:
When using this toolbox today, I found two corner cases: comparisons with infs and nans were not working since they are considered doubles. Better to differentiate then, since @isinf and @isnan exists.

I also forgot to include the joinCell function used in the test setup. This is part of CellUtils.

Util/Schema/createTree.m
ggalibert: I don't mind which way you want it to be done if you can confirm with the devs that the license doesn't need to be in every file header. However I think it's important to stick to a consistent template, whichever you think is best so there is probably a separate task of harmonising them at some stage to a new template, but in the meantime probably preferable to keep consistency with current template?

ocehugo: just checked on GPL stuff - it is required on every file. I thought it was the same as MIT (just put a license file and all good). The copyright should also be in every file. 

PS: This kinda of information should be somewhere in a "contributing" section. Probably also worth to mention in the PR template (I will update the template that I made here).

ocehugo:
@ggalibert , all docstrings/licenses in.

PR #583, opened 2019-09-13
include StructUtils functions

ocehugo:
This is a req. for the generic readers.


ocehugo:
just did the same as in CellUtils branch..

PR #584, opened 2019-09-13
import of cellUtils funcs

ocehugo:
This is required for Schema testing and Generic
Header/Data readers.

### DIFF
--- a/Util/CellUtils/cell2str.m
+++ b/Util/CellUtils/cell2str.m
@@ -1,0 +1,67 @@
+function [cstr] = cell2str(xcell, sep)
+    % function [cstr] = cell2str(xcell,sep)
+    %
+    % Convert all items in a cell to string and concatenate
+    % then with a separator, ignoring non-numeric entries.
+    %
+    % Inputs:
+    %
+    % xcell - a cell
+    % sep - a string separator - default to ','
+    %
+    % Outputs:
+    %
+    % cstr - a string of all strings and numerics separated by
+    %        `sep`.
+    %
+    % Example:
+    % cstr = cell2str({1,'string'})
+    % assert(strcmp(cstr,'1,string'))
+    %
+    % author: hugo.oliveira@utas.edu.au
+    %
+
+    % Copyright (C) 2019, Australian Ocean Data Network (AODN) and Integrated
+    % Marine Observing System (IMOS).
+    %
+    % This program is free software: you can redistribute it and/or modify
+    % it under the terms of the GNU General Public License as published by
+    % the Free Software Foundation version 3 of the License.
+    %
+    % This program is distributed in the hope that it will be useful,
+    % but WITHOUT ANY WARRANTY; without even the implied warranty of
+    % MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    % GNU General Public License for more details.
+    %
+    % You should have received a copy of the GNU General Public License
+    % along with this program.
+    % If not, see <https://www.gnu.org/licenses/gpl-3.0.en.html>.
+    %
+
+    if nargin < 2
+        sep = ',';
+    end
+
+    if ~ischar(sep)
+        error('second argument `sep` must be string');
+    end
+
+    cn = numel(xcell);
+    cstr = '';
+    ksep = '';
+
+    for k = 1:cn
+        data = xcell{k};
+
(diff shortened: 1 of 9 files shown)


Util/CellUtils/cell2str.m
```diff
@@ -0,0 +1,67 @@
...
+
+    for k = 1:cn,
+        data = xcell{k};
+
+        if isnumeric(data),
+            cstr = [cstr ksep num2str(data)];
+        elseif isstr(data),
+            cstr = [cstr ksep data];
+        else
+            disp('Ignored entry %d in cell', k)
```
ggalibert: I've tried to uniform the format of any message displayed so that a user knows whether this is info, warning or error. See  `./AutomaticQC/imosSideLobeVelocitySetQC.m` for example.
```Matlab
disp(['Info : imosSideLobeVelocitySetQC uses nominal depth because no pressure data in file ' sample_data.toolbox_input_file]);
```

Util/CellUtils/joinCell.m
```diff
@@ -0,0 +1,75 @@
...
+    c1_not_row = c1size(2) ~= 1;
+    c2_not_column = c2size(1) ~= 1;
+    c2_not_row = c2size(2) ~= 1;
+
+    invalid_c1 = c1_not_column && c1_not_row;
+    invalid_c2 = c2_not_column && c2_not_row;
+    invalid_input = invalid_c1 || invalid_c2;
+
+    if invalid_input,
+        error('Cell is not a vector.')
```
ggalibert: Try to always end you line with ';'. Matlab has an embedded code checker where something not optimal will be underlined in orrange. Hovering it gives a hint on how to write better code. In this case the '=' sign should be underlined. In general try to pay attention to this feedback from the Matlab editor, this can be very useful sometimes.

Util/CellUtils/popFromCell.m
```diff
@@ -0,0 +1,79 @@
...
+    %
+
+    csize = size(xcell);
+    flen = prod(csize);
+    newcell = cell(1);
+
+    not_cell_or_str = not(iscell(match) || ischar(match));
+
+    if not_cell_or_str,
+        error('match need to be cell or str')
```
ggalibert: ends with ';'

Util/CellUtils/cell2str.m
ocehugo: Thanks - the line is actually wrong, disp only accepts one argument.

Util/CellUtils/joinCell.m
ocehugo: My matlab editor/linter complains about other stuff but not about the missing ; in error calls. I'm resolving some warnings anyway

Util/CellUtils/popFromCell.m
ocehugo: same as above

Util/CellUtils/cell2str.m
ocehugo: I provided a new function - swarning - in the commit below. It takes care of knowing the scope so we don't need to provide the name of the function as a string in the argument. The function reports the full scope in the call chain (a la error).

ocehugo:
most warnings clear from mlint and other small fixes.

Util/CellUtils/cell2str.m
```diff
@@ -38,27 +38,27 @@
...
 
-        if isnumeric(data),
+        if isnumeric(data)
             cstr = [cstr ksep num2str(data)];
-        elseif isstr(data),
+        elseif ischar(data)
             cstr = [cstr ksep data];
         else
-            disp('Ignored entry %d in cell', k)
+            swarning('Ignoring entry %d in first argument', k);
```
ggalibert: Is this a Matlab function?

Util/swarning.m
```diff
@@ -0,0 +1,82 @@
+function swarning(varargin)
```
ggalibert: This should be included in one of the relevant commit!

ocehugo: done, made into the first commit of import functions

Util/CellUtils/cell2str.m
ocehugo: done, see rewritten history.

ocehugo:
This is ready to merge, but don't merge since I have to fix the previous merge that was not squashed.

PR #585, opened 2019-09-25
repo-wise eol normalization

ocehugo:
This commit configure the repo to handle crlf/non-crlf
line endings and also normalized the line-endings
in most source files of the repo to be non-crlf.

The modified files here are "crlf" with windows style
end-of-line markers. They are all changed to non-crlf
eol markers (unix), given this is the majority of the files.

New files (crlf or not) are handled automatically from
this commit onwards.

For compatibility reasons, instrument data files, java code
and coefficients files are kept as is (assumed binary format),
while .bat files are kept crlf.


ocehugo:
guys, you can pass your eye on, but it will be boring - just changing line-endings in windows written files.

PR #589, opened 2019-09-29
Runtime v96 support, closes #588

### LINKED ISSUE #588: Binary deployments broken with matlab_runtime v96
Updating the matlab_runtime to v96 broke the binary GUI.

The problem was that in v96 the path of some images was changed.

- [x] Update paths in GUI code to match the ones in the new runtime.

ocehugo:
Correct paths for the new v96 Matlab runtime environment.

I also add a correction to the new build script since the old java files were not being replaced and were causing missing classes when opening the GUI.


ocehugo:
@lbesnard , can you approve please? Just merged master to fix conflicts.

PR #581, opened 2019-09-11
[WIP] Same cphl var, closes #532

ocehugo:
This includes:

- [x] removal CPHU/CPHF
- [x] ensure CPHL is enumerated when more than one CPHL variable is defined.
- [x] provide some small clean up in some files (linter).
- [x] ensure comments are equal in all readers
- [ ] Tests for all affected readers.

Two bits:
1. I went forward and standardized the comments for all CPHL. Legit?
2. I put unknown coefficients for the majority of the cases, please advise if this should be changed.

PR #590, opened 2019-10-01
Fixed cphl same var, closes #532

### LINKED ISSUE #532: CPHL, CHLF and CHLU are the same parameter with different attributes
CPHL, CHLF and CHLU should be made distinct in the comment variable attribute:

CPHL: conversion from fluorescence to chlorophyll-a using uknown coefficients (most likely factory ones though, would be good to confirm).
CHLF: conversion from fluorescence to chlorophyll-a using factory coefficients.
CHLU: conversion from fluorescence to chlorophyll-a using user coefficients.

ocehugo:
This is #581 rebased with the fixes.

Please merge asap

This includes:

- [x] removal CPHU/CPHF
- [x] ensure CPHL is enumerated when more than one CPHL variable is defined.
- [x] provide some small clean up in some files (linter).
- [x] ensure comments are equal in all readers and excitation/scattering comments are kept.
- [x] Tests for all affected readers provided
- [x] tests shows that all readers affected can read data with no CPHU/CPHF created.


ggalibert:
As usual, commits are not in best order. I would have removed CHLU/CHLF from parsers prior to removing them from config files.

Careful, check that code is CPHL, not CHPL or anything else!

Parser/readWQMdat.m
```diff
@@ -1,438 +1,440 @@
...
+    params{end + 1} = {'HH:MM:SS', {'', ''}};
+    params{end + 1} = {'Cond(mmho)', {'CNDC', ''}}; % mmho <=> mS , I think /mm is assumed
+    params{end + 1} = {'Cond(S/m)', {'CNDC', ''}};
+    params{end + 1} = {'Temp(C)', {'TEMP', ''}};
+    params{end + 1} = {'Pres(dbar)', {'PRES_REL', ''}};
+    params{end + 1} = {'Sal(PSU)', {'PSAL', ''}};
+    params{end + 1} = {'DO(mg/l)', {'DOXY', ''}};
+    params{end + 1} = {'DO(mmol/m^3)', {'DOX1', ''}}; % mmol/m3 <=> umol/l
+    params{end + 1} = {'DO(ml/l)', {'DOX', ''}};
+    params{end + 1} = {'CHL(ug/l)', {'CPHL', getCPHLcomment('unknown','470nm','695nm')}}; %older files
```
ggalibert: I would prefer you include all function declaration that you introduce in the same commit. This is OK if we squash but not easy to review!

ggalibert: In addition this is inconsistent with other parsers where you don't use this function.

Parser/YSI6SeriesParse.m
```diff
@@ -1,323 +1,329 @@
...
-    field = records.(fields{k});
-
-    % dimensions definition must stay in this order : T, Z, Y, X, others;
-    % to be CF compliant
-    if any(strcmpi(field, {'LATITUDE', 'LONGITUDE'}))
-        sample_data.variables{end+1}.dimensions = [];
-    else
-        sample_data.variables{end+1}.dimensions = 1;
+function sample_data = YSI6SeriesParse(filename, mode)
+    %YSI6SERIESPARSE Parser for YSI 6 series MultiParameter data logger files.
```
ggalibert: Again impossible to review if you change the whole indent in file at the same time you're making other changes.

Parser/YSI6SeriesParse.m
```diff
@@ -1,323 +1,329 @@
...
+    % read in the whole file into 'data'
+    fid = -1;
+
+    try
+        fid = fopen(filename, 'rb');
+        data = fread(fid, inf, '*uint8');
+        fclose(fid);
+    catch e
+        if fid ~= -1, fclose(fid); end
+        rethrow(e);
```
ggalibert: This commit is about converting any CHLF/CHLU into CPHL but you've made other changes. It would be clearer if they were in another commit and you explainwhat are the changes for this commit.

ocehugo:
> As usual, commits are not in best order. I would have removed CHLU/CHLF from parsers prior to removing them from config files.


No probs, I can reorder 

> 
> Careful, check that code is CPHL, not CHPL or anything else!

there is no CHPL, only CPHL

Parser/YSI6SeriesParse.m
ggalibert: If you want to change the indent because you think it's not correct, that's fine but that's a separate task where you could do this across the whole repo in one commit.

Parser/readWQMraw.m
```diff
@@ -620,8 +620,8 @@
     params{end + 1} = {'DO(mmol/m^3)', {'DOX1', ''}}; % mmol/m3 <=> umol/l
     params{end + 1} = {'oxygen', {'DOX', ''}};
     params{end + 1} = {'fluorescence', {'FLU2', ''}};
-    params{end + 1} = {'F_Cal_CHL', {'CHPL', getCPHLcomment('factory','470nm','695nm')}};
-    params{end + 1} = {'U_Cal_CHL', {'CHPL', getCPHLcomment('user','470nm','685nm')}};
+    params{end + 1} = {'F_Cal_CHL', {'CHPL', getCPHLcomment('factory', '470nm', '695nm')}};
```
ggalibert: CHPL here!!! Should be CPHL.

Parser/readWQMraw.m
```diff
@@ -568,296 +593,316 @@
...
+    params{end + 1} = {'conductivity', {'CNDC', ''}};
+    params{end + 1} = {'temperature', {'TEMP', ''}};
+    params{end + 1} = {'pressure', {'PRES_REL', ''}};
+    params{end + 1} = {'salinity', {'PSAL', ''}};
+    params{end + 1} = {'DO(mg/l)', {'DOXY', ''}};
+    params{end + 1} = {'DO(mmol/m^3)', {'DOX1', ''}}; % mmol/m3 <=> umol/l
+    params{end + 1} = {'oxygen', {'DOX', ''}};
+    params{end + 1} = {'fluorescence', {'FLU2', ''}};
+    params{end + 1} = {'F_Cal_CHL', {'CHPL', getCPHLcomment('factory', '470nm', '695nm')}};
+    params{end + 1} = {'U_Cal_CHL', {'CHPL', getCPHLcomment('user', '470nm', '685nm')}};
```
ggalibert: Same typo here.

Util/getCPHLcomment.m
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    %TODO this should be build on a table for each instrument...
+    chla_str = ['Artificial chlorophyll data ' ...
+                'computed from bio-optical sensor raw counts measurements. The ' ...
+                'fluorometre is equipped with a ' excitation_wl_str ' peak wavelength ' ...
+                'LED to irradiate and a photodetector paired with an optical filter ' ...
+                'which measures everything that fluoresces in the region ' ...
+                e_conn ' ' scattered_wl_str '. ' ...
+                'Originally expressed in ug/l, 1l = 0.001m3 was assumed.'];
```
ggalibert: Are you sure this is always the case for original unit?

Parser/convertSBEcnvVar.m
```diff
@@ -137,9 +140,11 @@
                 comment = 'PAR/Logarithmic/ Satlantic';
 
                 % artificial chlorophyll from fluorescence (mg/m3)
+                % SBE manual indicate this is a Wet Labs - ECO AFL sensor
             case 'flECO0x2DAFL'
                 name = 'CPHL';
-                comment = 'Artificial chlorophyll data computed from bio-optical sensor raw counts measurements.';
```
ggalibert: In this case for example the unit is already in the right form. It wasn't in ug/l. Please re-check across all parsers.

test/testOnlyCPHL.m
```diff
@@ -0,0 +1,248 @@
...
+        end
+
+        function test_sbe9p(testCase, sbe9p_reader, sbe9p_v000_param, mode)
+            data = sbe9p_reader({sbe9p_v000_param}, mode);
+            assert(check_cphl(data));
+        end
+
+        function test_wqmraw(testCase, wqmraw_reader, wqmraw_v000_param, mode)
+            data = wqmraw_reader(wqmraw_v000_param, mode);
+            assert(check_cphl(data));
```
ggalibert: This test should have picked up the CHLP typo?!

Parser/readWQMdat.m
ocehugo: Correct - this was not supposed to be here. 
I probably merged something when doing the rebase for the WQMs files.

Parser/YSI6SeriesParse.m
ocehugo: The problem here was that the file was not indented and the default diff algorithm/options created a confused diff. If using "-w" you will see that the only differences are newlines between flow statements (for,if, etc).

Parser/YSI6SeriesParse.m
ocehugo: as above, the diff is confusing because of default options and the indents. There is no change.

Parser/readWQMraw.m
ocehugo: Ahh, again, the same thing - this came through WQM merge at wrong place.

Parser/readWQMraw.m
ocehugo: as above

Util/getCPHLcomment.m
ocehugo: Yes. All comment are the same in all files that mention fluorescence. Unless you want to change everything to "mg/m3" instead?

Parser/convertSBEcnvVar.m
ocehugo: According to the SBE manual, flECOAFL is outputed in mg/m3 which is ug/l, so it's correct.

test/testOnlyCPHL.m
ocehugo: If you run the test at the commit it was introduced, it should pass.

ocehugo:
see #594

PR #602, opened 2019-10-22
Release 2.6.1

ocehugo:
What this is about: Add some gitignores, rever a commit that broke the stand-alone Linux app, remove deprecated Win32 app/scripts, include new Java ddb, new Windows/Linux stand-alone apps.

All stand-alone apps and the toolbox within MATLAB are able to load files, run some preprocessing routines, present the main window, open the metadata tab and export a NetCDF file.

Please merge asap


lbesnard:
I wanted to test the linux version. I followed the wiki 
https://github.com/aodn/imos-toolbox/wiki/ToolboxInstallation
I think it would be great to specify in the wiki the direct link to the MCR for both Linux and Win
It's not clear to me that ```glnxa64``` is for linux and that this is the string I have to look for in the filename.

I also think the installation page for the wiki should be divided between a Linux and windows installation. This would be heaps clearer.

Anyhow, I followed the instructions, added the MCR to the path. I get an error
```
./imosToolbox_Linux64.bin: error while loading shared libraries: libmwlaunchermain.so: cannot open shared object file: No such file or directory

```
I tried following the suggestions in that [page](https://au.mathworks.com/matlabcentral/answers/267562-how-can-i-resolve-this-mcc-runtime-error-cannot-open-shared-object-library) but still get an error
```
./imosToolbox_Linux64.sh -h /opt/MATLAB/MATLAB_Runtime/v95/runtime/glnxa64
------------------------------------------
Setting up environment variables
---
LD_LIBRARY_PATH is .:-h/runtime/glnxa64:-h/bin/glnxa64:-h/sys/os/glnxa64:-h/sys/java/jre/glnxa64/jre/lib/amd64/native_threads:-h/sys/java/jre/glnxa64/jre/lib/amd64/server:-h/sys/java/jre/glnxa64/jre/lib/amd64/client:-h/sys/java/jre/glnxa64/jre/lib/amd64
./imosToolbox_Linux64.bin: error while loading shared libraries: libmwlaunchermain.so: cannot open shared object file: No such file or directory
```
I'm sure this is not related to that PR, however I can't run it following the instructions.

Last, as a ref for @ggalibert, this PR is from a branch sitting on @ocehugo 's repo, so to checkout the branch from a local copy of the aodn repo, you have to follow those steps
https://stackoverflow.com/questions/5884784/how-to-pull-remote-branch-from-somebody-elses-repo

ocehugo:
In short : to use stand-alone, you have to use the .sh script. Your call is wrong, just remove the -h and it should work.

Will have a look at the wiki to see what can be improved and provide a more straightforward guide then

ocehugo:
I just update the wiki, hopefully is more straight forward

PR #614, opened 2019-11-18
Filenames with deployment field, closes #541

### LINKED ISSUE #541: filename date start and end should match time_deployment_start/end attributes
> The revised dt_start and dt_end columns are now published at http://oceancurrent.imos.org.au/timeseries/ with a note headed 26 April, saying "There are now more zeros in those columns, indicating correspondence of the dates in the filenames with either the range of the time vector, the period of good data, or both." I think we need to clarify the purpose of those dates in the filenames (and make sure all are the same format - that's what triggered the error).
> -- @DavidGriffin1

ocehugo:
This move the filename conventions to use the ```time_deployment``` fields isntead of ```time_coverage``` fields.

@ggalibert , please check if this is what you had in mind in #541

### DIFF
--- a/IMOS/genIMOSFileName.m
+++ b/IMOS/genIMOSFileName.m
@@ -171,7 +171,7 @@ function config = genDefaultFileNameConfig(sample_data, dateFmt, mode)
       dateFmt = dateFmt(1:end-1);
       extraChar = 'Z';
   end
-  config.start_date    = [datestr(sample_data.time_coverage_start, dateFmt), extraChar];
+  config.start_date    = [datestr(sample_data.time_deployment_start, dateFmt), extraChar];
   
   config.site_code     = sample_data.site_code;
 
@@ -197,7 +197,7 @@ function config = genDefaultFileNameConfig(sample_data, dateFmt, mode)
   config.product_type(config.product_type == '_') = '-';
   
   % <end_date>, <creation_date>
-  config.end_date    = [datestr(sample_data.time_coverage_end, dateFmt), extraChar];
+  config.end_date    = [datestr(sample_data.time_deployment_end, dateFmt), extraChar];
   
   config.creation_date = [datestr(sample_data.date_created, dateFmt), extraChar];
 end


ocehugo:
@ggalibert, I think @lbesnard is waiting your :+1: on that to merge

IMOS/finaliseData.m
```diff
@@ -161,47 +161,28 @@
...
-              if isempty(sam.time_coverage_end),
-                  sam.time_coverage_end   = sam.dimensions{iTime}.data(end);
-              end
-          else
-              if isempty(sam.time_coverage_start), sam.time_coverage_start = []; end
-              if isempty(sam.time_coverage_end),   sam.time_coverage_end   = []; end
-          end
-  
+  % set the time coverage/deployment period from the data
+  iTime = getVar(sam.variables, 'TIME');
```
ggalibert: I disagree with this refactoring. In profile mode TIME is in sam.variables while in timeseries mode TIME is in sam.dimensions. 
If you really want to refactor you could declare a string modeStr that you set either to 'variables' or 'dimensions' depending on the mode:
```Matlab
switch mode
  case 'profile'
    modeStr = 'variables';
  case 'timeSeries'
    modeStr = 'dimensions';
end

iTime = getVar(sam.(modeStr), 'TIME')
```
 ... etc

IMOS/finaliseData.m
```diff
@@ -161,47 +161,28 @@
...
+      sam.time_deployment_start = sam.variables{iTime}.data(1);
+    end
+    if isempty(sam.time_deployment_end)
+      sam.time_deployment_end = sam.variables{iTime}.data(end);
+    end
+  else
+    if isempty(sam.time_coverage_start), sam.time_coverage_start = []; end
+    if isempty(sam.time_coverage_end), sam.time_coverage_end   = []; end
+    if isempty(sam.time_deployment_start), sam.time_deployment_start = []; end
+    if isempty(sam.time_deployment_end), sam.time_deployment_end = []; end
```
ggalibert: You don't initialise the `_origin` attributes, did you check this doesn't have a negative impact?
Also, you changed the logic in comparison to previous code, are you sure this is needed? This was not needed to tackle https://github.com/aodn/imos-toolbox/issues/541

IMOS/finaliseData.m
ocehugo: I forgot that - good catch. It would be nice to fix this mess-up - TIME is always a variable so it should only be store in the variable fieldname. I shall raise a small refactoring issue for that.

I implemented the fix with ```if``` so it is shorter and more explicit. Tests are also added - completely forgot to run/include them.

IMOS/finaliseData.m
ocehugo: thanks - I'll add them in the 'deployment' if block above which is where they should be.

The reason why I changed the logic from the previous code was that were duplicate checks for the same thing several times.

ocehugo: @ggalibert , in fact, _origin attributes are initialized by makeNEtCDFCompliant to an empty string. FinaliseData.m overwrite this to an empty array, which did not make any sense, since they are set to strings.
Hence, if there is ever a need to append information to the origin attribute, the current functionality already handle it (non empty values are strings).

This is why lines were deleted - I will update the commit msg.

ocehugo: restore previous logic - too hard to test.

ocehugo:
@lbesnard , this became a very simple commit since I removed some refactorign not needed.

ocehugo:
@ggalibert - there is only a case here that may break a workflow - for very simple processing streams (no ddb and no PP/QC runs), the filenames only get creation dates since time_deployment is empty (but not time_coverage).

I'm just wondering if this is someone workflow outside of us simply testing the export to netcdf functionality.

PR #616, opened 2019-11-19
feat(Preprocessing): allow multi var names in aquatrackapp

ocehugo:
This commit remove previous warning message and put
a simple logic handler for multi vars in the
aquatracka pre-processing function.

The multi-var case should be quite rare but happens
during tests since we create fake voltage variables in
structures with CPHL data already.

### DIFF
--- a/Preprocessing/aquatrackaPP.m
+++ b/Preprocessing/aquatrackaPP.m
@@ -94,10 +94,16 @@ for k = 1:length(sample_data)
       else
           coordinates = '';
       end
-    
-      if inCell(all_current_names,imosAquatracka{i}) 
-        %not suppose to happen, but warn if does since this will probably lead to overwrite at the file level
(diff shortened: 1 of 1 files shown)


lbesnard:
@ocehugo in case of a multivar case, is there a variable attribute added somewhere to help users know what are the differences between *_1 and *_2 ? or is this kind of information just not available?

ocehugo:
@lbesnard,

short answer: this is handled by the comment attribute.

long answer:
This function only adds **one** chla variable. The problem is to find the correct name to avoid name clashes, since the receiving structure may already have ```CPHL```, or ```CPHL_2```, etc.

The above code just make sure that, if we got any ```CPHL_[N]``` already, the preprocessed variable will be stored as ```CPHL_[N+1]``` with the proper  comment (located at  ```commentsAquatracka``` variable)

PR #619, opened 2019-11-25
feat(path): include a basename function

ocehugo:
A function that mimic the gnu coreutils basename function


ocehugo:
@lbesnard, I just did a re-implementation for windows support. The function is here to exactly avoid writing the logic cases everytime.

PR #623, opened 2019-12-13
Nortek signature fixes, closes #554

### LINKED ISSUE #554: Sampling regime metadata set incorrectly for Signature ADCP
We have received [this file](https://github.com/aodn/imos-toolbox/files/3323062/IMOS_ANMN-WA_AETVZ_20181203T080004Z_WACA20_FV01_WACA20-1812-Signature250-194_END-20190620T023004Z_C-20190624T043041Z.nc.zip), from a "Norrtek Signature250" instrument, with the following global attributes:
```
                :instrument_burst_duration = 0. ;
                :instrument_burst_interval = 900. ;
                :instrument_sample_interval = -17173800. ;
```
The timestamps are at regular 15min intervals, so if it was originally burst sampled, the bursts have already been binned. In that case the `instrument_burst_` attributes should not be present (or at least set to correct values).

Either way, the `instrument_sample_interval` should be set to a sensible value (in this case 900.).

ocehugo:
See commit msgs.

### DIFF
--- a/Parser/GenericParser/DataParser.m
+++ b/Parser/GenericParser/DataParser.m
@@ -1,0 +1,207 @@
+function [rawdata, data] = DataParser(fid, sdata)
+% function [rawdata,data] = DataParser(fid,sdata)
+%
+% A wrapper function to parse data from a
+% file id with a pre-defined super data structure.
+% See Parsers and rules.
+%
+% Inputs:
+%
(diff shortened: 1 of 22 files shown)


lbesnard:
I'll have to trust you on that one. code looks good but it's very specific and I'm not sure I understand all the bells and whistles

Issue #549, opened 2019-05-31
starmon mini header changed


petejan:
StarODDI have changed the file header format to

```
#B	Created:	31/05/2019 11:51:00
##	Version	SeaStar	7.91
#5	Field separation:	0
#2	Date & Time:	1
#6	Decimal point:	.
#7	Date def.:	dd/mm/yyyy	/
#8	Time def.:	:
##	Axis	0	Temperature(°C)	clRed	FALSE
##	Series	0	Temp(°C)	clRed	0	Temp(°C)	2
##	Recorder	9	Starmon mini	4035
##	Chart	9T4035DAT	
#D	Data:	69802	01/08/2018 00:00:00	31/03/2019 08:45:00
1	01/08/2018 00:00:00	9.69
2	01/08/2018 00:05:00	9.65
3	01/08/2018 00:10:00	9.61
4	01/08/2018 00:15:00	9.59

```
The parser and Wiki needs updating to head files generated by the newer version of software

ocehugo:
Guys,

this should be a change in place or a new reader option? For example, older files would still require the "old" reader right?

@petejan,
may you please provide the function call signature and an example file (old and new)!?


Requirements:
- [ ] new header reader is implemented
- [ ] old/new reader is selectable
- [ ] Tests for new reader is available

ggalibert:
I suppose any old and new formats should be supported.

petejan:
Attached example files from several sensors I have,

[starODDI-example.zip](https://github.com/aodn/imos-toolbox/files/3415169/starODDI-example.zip)

ocehugo:
@petejan,

For completeness, our current test file repo is missing some Staroddi example files:

a. Star-oddi with temperature corrections 
b. Star-oddi with temperature in degF
c. Star-oddi with pressure corrections
d. Star-oddi with salinity data.
e. Star-oddi with pressure correction and salinity data.

any mini/dst-tilt/dst-ctd would do.

So far I have faked the above conditions based on the manual but it would be nice to have real files.

would you have any to provide to us?

Cheers

petejan:
Attached data files from temperature sensors, conductivity, and pitch and roll senors,

[starmon.zip](https://github.com/aodn/imos-toolbox/files/3954074/starmon.zip)

Remember the date format changes based on the users PC settings, also the data export from the starmon software needs setting up correctly.

ocehugo:
[...]

PR #628, opened 2019-12-18
feat(Parsers): Re-use old parser names to new parser, closes #622 #549

### LINKED ISSUE #622: StarmonMiniParse.m file header formats changed
I have the following header formats in our latest Starmon Mini files: (Ignore the apostrophies at the start, put them in to stop auto markdown, screenshot for clarity)

Screen Shot 2019-12-12 at 4 18 18 pm

'#B	Created:	25.09.2019 03:52:48
'##	Version	SeaStar	7.61
'#5	Field separation:	0
'#2	Date & Time:	1
'#6	Decimal point:	,
'#7	Date def.:	dd.mm.yyyy	.
'#8	Time def.:	:
'##	Axis	0	Temperature(?C)	clRed	FALSE
'##	Series	0	Temp(?C)	clRed	0	Temp(?C)	2
'##	Recorder	3	Starmon mini	3998
'##	Chart	3T3998DAT	
'#D	Data:	152105	15.04.2018 00:00:00	25.09.2019 03:20:00

The headerFormat in getHeader subfunction doesn't work with these headers:

headerFormat = '#%d\t%[^:]:\t%[^\r\n]';
headerContent = textscan(fileId, headerFormat, 'Delimiter', '');

Is it possible to get the format specifier updated soon? I can have a go if not!
Thanks

ocehugo:
This is a requirement for transparent usage of the
new Staroddi Parser that supports Mini v0, Mini v1, DST Tilt
and DST CTD.

The reason why is done this way is to still support
old insturments.txt configuration files, database
key:value associations, and to be transparent to the user.

### DIFF
--- a/Parser/StarmonDSTParse.m
+++ b/Parser/StarmonDSTParse.m
@@ -38,8 +38,8 @@ function sample_data = StarmonDSTParse( filename, mode )
 try
     sample_data = StaroddiParser(filename,mode);
     return
-catch
-
+catch e
+    warning("Using the old StarmonDSTParse. Please raise an issue in our github page with this warning msg: %s", e.message)
 end
 
 %TODO remove old code below
--- a/Parser/StarmonDSTParse.m
+++ b/Parser/StarmonDSTParse.m
@@ -35,6 +35,14 @@ function sample_data = StarmonDSTParse( filename, mode )
 % along with this program.
 % If not, see <https://www.gnu.org/licenses/gpl-3.0.en.html>.
 %
+try
+    sample_data = StaroddiParser(filename,mode);
+    return
+catch
+
+end
+
+%TODO remove old code below
 narginchk(1,2);
 
 if ~iscellstr(filename)
--- a/Parser/StarmonMiniParse.m
+++ b/Parser/StarmonMiniParse.m
@@ -35,6 +35,14 @@ function sample_data = StarmonMiniParse( filename, mode )
 % along with this program.
 % If not, see <https://www.gnu.org/licenses/gpl-3.0.en.html>.
 %
+try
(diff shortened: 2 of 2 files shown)


Parser/StarmonDSTParse.m
```diff
@@ -35,6 +35,14 @@
 % along with this program.
 % If not, see <https://www.gnu.org/licenses/gpl-3.0.en.html>.
 %
+try
+    sample_data = StaroddiParser(filename,mode);
+    return
+catch
+
+end
```
lbesnard: is this really the best way to do this? bypassing the entire code? couldn't this be done in the function calling this parser?
otherwise, it would be great to add some comment above the try catch about the reasons why all the rest of the code is bypassed. And maybe a printf or something in after the catch to mention the default parser is used?

ocehugo: > Is this really the best way to do this? bypassing the entire code? couldn't this be done in the function calling this parser?

It's the most transparent way for the end-users. Users and their configurations (toolbox and db's) are sometimes hardcoded for the older parser names. Moving to a different name would not allow backward compatibility - I discussed yesterday with gg.

>t would be great to add some comment above the try catch about the reasons why all the rest of the code is bypassed. And maybe a printf or something in after the catch to mention the default parser is used?

Yes, I'll put a msg on the exception. I was thinking that the user should not be aware of the fail but if we don't print anything we will never know if someone had a problem with the new parser.

The right way is to completely remove the entire code outside the try and just fail miserably.  Although we got a good test coverage for this parser and the old parser is pretty much deprecated, I'm just fencing a bit high here until q3 release.

ocehugo:
@lbesnard done

PR #644, opened 2020-01-23
Add burst_duration and burst_interval

sspagnol:
Copy across instrument_burst_duration and instrument_burst_interval to meta struct.


ocehugo:
I will just hold this until i write a test function for the parser.

PR #656, opened 2020-04-08
Spike qc nonburst - part 1

ocehugo:
Hey guys,

this is the code for the Spike QC functionality that need to be merged for the Q3 release.

What's included:

- Some small cleanups
- Some refactoring/plumbing in previous routines to be more robust and permit burst processing in the future (as well as extra checks regarding time variables).
- The actual Spike Routines (Hampel,SavGol,etc).
- The SpikeQC driver with UI elements.

I poking @ggalibert since he will be in a better position to review the UI elements I think. It's pretty spartan atm, but I seek to improve it for the next spike processing (bursts). Most important thing to be done yet is to easy allow the users to change the filtering parameters. I tried my best to code some dynamical UI textboxes, but I didn't have enough time to finish and debug the callback hell :)


Util/StructUtils/combineStructFields.m
```diff
@@ -5,20 +5,37 @@
...
+    % assert(combined.y==3);
+    %
+    % % multi structure case
+    % a = struct(); b=a; c=a;
+    % a.x = 1;
+    % b.x = 2;
+    % b(2).x = 3;
+    % b(3).x = 4;
+    % c.x = 3;
+    % c(2).x = 4;
```
lbesnard, line 32: So I'm confused about this structure combination. It seems like the order of arguments is important. Why combined(2) == 4 (from the c value) and not  3 (from the b value) ? is it really what you want?

Util/Time/stepsInDay.m
```diff
@@ -0,0 +1,55 @@
...
+% tunit - string representing a time/precision unit.
+%
+% Outputs:
+%
+% steps - a floating point number
+%
+% Example:
+%
+% [steps] = timeQuantisationStep('minute')
+% assert(isequal(step,1440);
```
lbesnard, line 18: step with an s

AutomaticQC/imosSpikeQC.txt
```diff
@@ -0,0 +1,30 @@
+% comma separated variable names/subnames
+allowed_variables = CDOM, CNDC, CPHL, OXY, PAR, PSAL, SST, TEMP, TURB
```
lbesnard, line 2: why only this set of variables? Is there a reason we're not doing this on the whole set ?
https://github.com/aodn/imos-toolbox/blob/master/IMOS/imosParameters.txt

Util/StructUtils/combineStructFields.m
ocehugo: Yes, order precedence is exactly what I seek here. This example is here to actually point to this behaviour. The function is "equivalent" of dict.update() in python. It will "append with overwrite" all the arguments fields.

In another words:
```combined(2).x == 4``` because the last structure holding the ```x``` field at position ```(2)``` is the c struct.

Another example:
```
b(4).x = 100;
combined = combineStructfields(c,b,a)
assert(isequal(combined.x,[1,3,4,100]))
```

Util/Time/stepsInDay.m
ocehugo: Yeap - function name is wrong too, as well as assertion line.

AutomaticQC/imosSpikeQC.txt
ocehugo: yeap, fencing on purpose for several reasons:

1. I don't want people applying this to untested variables,  ATM.
2. I didn't code the cases for two or more dimensional variables (e.g. velocity).
3. I got a very polluted UI when there was a lot of variables because I had to add text + dropdown menu + buttons. This was dragging me down too. Hence, I fenced down the vars to the most used/ful ones and simplified the interface.

The idea is to expand this, but only after I managed a feasible UI design that allows users to quickly swap values/methods for different variables for example.

ocehugo:
>@ocehugo I think it would be quite valuable to add some unittest on the spikeQC function/wrappers to check the qc output on a given spiky array. It would help us validating the migration to a different version of matlab as well.
What do you think?

Agreed. However, the tests can only be done at the functional level and not UI level. I got a half-baked suite here dangling with a mix of basic tests and realistic ones (from netcdfs) for the hampel method. I'll do some clean up, mix/match the tests among the available tests,  and put it here. Some realistic tests I will leave for later because I don't want to delay this anymore and I also need to cherry-pick other variables.

The challenge here is to pick-up some good cases apart from artificial ones

ocehugo:
comments addressed in the new commits @lbesnard

ocehugo:
@lbesnard  - I just add some extra commits fixing stuff that I encounter when trying to build the standalone packages and further issues when trying to mint the release.

PR #666, opened 2020-06-16
Update tilt aquadopp, closes #662

### LINKED ISSUE #662: Update aquadoppdw cutoff tilt angle
From #635

Requirements:
- [ ] Include Tilt Values for AquadoppDWs

ocehugo:


AutomaticQC/imosTiltVelocitySetQC.m
```diff
@@ -241,27 +267,55 @@
...
+  secondTiltflag = params{5}(k);
+  return
 end
 
-end
\ No newline at end of file
+%finally, try to match specific options on the table as substrings in the instrument
+%name - useful when loading from netcdfs
+for k=1:nlen
+  invmatch = contains(instrument,names{k},'IgnoreCase',true);
```
lbesnard, line 310: further below, I see there are both long ranger and longranger (not from your PR though).
I have the feeling IgnoreCase should be true, but also a "IgnoreWhitespaces' or something like this, some sort of fuzzysearchfinder

AutomaticQC/imosTiltVelocitySetQC.m
```diff
@@ -169,16 +171,29 @@
...
+if isempty(firstTiltThreshold) && auto
     % couldn't find this instrument so quit the test
     disp(['Warning: imosTiltVelositySetQC could not be performed on ' sample_data.toolbox_input_file ...
         ' instrument = "' instrument '" => Fill imosTiltVelositySetQC.txt with relevant make/model information if you wish to run this test on this dataset.']);
     return;
 end
 
+if ~auto
+  %fire-up a dialog box to confirm values/allow editing.
+  isdeg = @(x)(x>=0 & x<=360);
```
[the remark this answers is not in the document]
ocehugo: >sdeg as opposed to ? radians ?II don't see how you make sure the value is in degrees based on its value

Yes. We can't indeed, this is why the text of that box says ```firstTiltThreshold [deg]```.

AutomaticQC/imosTiltVelocitySetQC.m
ocehugo: Haha, we follow the same line of thought. I even did implement a Levenshtein distance for fuzzy matching and tested it, but it was not backward compatible with current entries.

lbesnard: ah thats a shame!

PR #669, opened 2020-07-07
SpikeQC Burst Spike functionality, Preview, and small fixes

ocehugo:
See commit msgs.

### DIFF
--- a/AutomaticQC/SpikeClassifiers/imosSpikeClassifierBurstHampel.m
+++ b/AutomaticQC/SpikeClassifiers/imosSpikeClassifierBurstHampel.m
@@ -71,18 +71,18 @@ function [spikes] = imosSpikeClassifierBurstHampel(bindexes, signal, use_burst_w
 narginchk(2, 7)
 
 if nargin == 2
-    use_burst_window = 1;
-    half_window_width = 0;
-    madfactor = 10;
+    use_burst_window = 0;
+    half_window_width = 1;
+    madfactor = 5;
     lower_mad_limit = 0;
     repeated_only = false;
 elseif nargin == 3
-    half_window_width = 0;
-    madfactor = 10;
+    half_window_width = 1;
+    madfactor = 5;
     lower_mad_limit = 0;
     repeated_only = false;
 elseif nargin == 4
-    madfactor = 10;
+    madfactor = 5;
     lower_mad_limit = 0;
     repeated_only = false;
 elseif nargin == 5
@@ -119,7 +119,7 @@ if ~isempty(bspikes)
     bs = bs + blen;
 end
 
-
+h=[];
 series_end_at_next_window = nbursts+1 >= length(bindexes)-nbursts;
 if series_end_at_next_window
     bie = bindexes{nbursts+1}(1);
--- a/AutomaticQC/SpikeClassifiers/imosSpikeClassifierBurstHampel.m
+++ b/AutomaticQC/SpikeClassifiers/imosSpikeClassifierBurstHampel.m
@@ -1,0 +1,176 @@
+function [spikes] = imosSpikeClassifierBurstHampel(bindexes, signal, use_burst_window, half_window_width, madfactor, lower_mad_limit, repeated_only)
+% function [spikes] = imosSpikeClassifierBurstHampel(bindexes, signal,  use_burst_window,  half_window_width, madfactor, lower_mad_limit, repeated_only)
+%
+% A Hampel classifier for burst data. The half_window_width is applied at burst level instead of
+% sample level.
+%
+% Inputs:
+%
+% bindexes - a cell with the invidivudal burst range indexes.
+% signal - A Burst 1-d signal.
+% use_burst_window  - a boolean to consider the half_window_width applied at burst scale. If false, will consider the burst series as continuous.
+% half_window_width - The half_window_width burst range (the full window size will be 1+2*half_window_width)
+% madfactor -  A multipling scale for the MAD.
+% repeated_only - a boolean to mark spikes in burst only if they are detected more than one time (only for half_window_width>0).
+% lower_mad_limit - a lower threshold for the MAD values, which values below will be ignored.
+%
+% Outputs:
+%
+% spikes - An array with spikes indexes.
+% fsignal - A filtered signal where the spikes are substituted by median values of the window.
+%
+% Example:
+%
+% % simple spikes
+% x = randn(1,100)*1e-2;
+% spikes = [3,7,33,92,99]
+% x(spikes) = 1000;
+% bduration = 6;
+% v = 1:bduration:length(x)+bduration;
+% for k=1:length(v)-1;
+% bind{k} = [v(k),min(v(k+1)-1,length(x))];
+% end
+% [dspikes] = imosSpikeClassifierBurstHampel(bind,x);
+% assert(isequal(dspikes,spikes));
+% % equal to Hampel
+% [dspikes] = imosSpikeClassifierBurstHampel(bind,x,length(bind));
+% [dspikes2] = imosSpikeClassifierHampel(x,length(bind));
+% assert(isequal(dspikes,spikes));
+% assert(isequal(dspikes,dspikes2));
+%
+% % detecting entire burst as spike
+% x = randn(1,100)*1e-2;
+% fullburst_spiked = [3,7,13,14,15,16,17,18,33,92,99]
+% x(fullburst_spiked) = 1000;
+% [dspikes] = imosSpikeClassifierBurstHampel(bind,x);
+% assert(isequal(dspikes,[3,7,33,92,99]));
+% [dspikes2] = imosSpikeClassifierBurstHampel(bind,x,true,2);
+% assert(isequal(dspikes2,fullburst_spiked))
+%
+%
+% author: hugo.oliveira@utas.edu.au
+%
+
+% Copyright (C) 2020, Australian Ocean Data Network (AODN) and Integrated
+% Marine Observing System (IMOS).
+%
+% This program is free software: you can redistribute it and/or modify
+% it under the terms of the GNU General Public License as published by
+% the Free Software Foundation version 3 of the License.
+%
+% This program is distributed in the hope that it will be useful,
+% but WITHOUT ANY WARRANTY; without even the implied warranty of
+% MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+% GNU General Public License for more details.
+%
+% You should have received a copy of the GNU General Public License
+% along with this program.
+% If not, see <https://www.gnu.org/licenses/gpl-3.0.en.html>.
+%
+
+narginchk(2, 7)
+
+if nargin == 2
+    use_burst_window = 1;
+    half_window_width = 0;
+    madfactor = 10;
+    lower_mad_limit = 0;
+    repeated_only = false;
+elseif nargin == 3
+    half_window_width = 0;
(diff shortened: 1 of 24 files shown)


AutomaticQC/SpikeClassifiers/imosSpikeClassifierOTSU.m
```diff
@@ -79,9 +79,9 @@
 end
 
 otsu = otsu_threshold(signal, nbins);
-threshold = otsu / oscale;
+threshold = otsu .* oscale;
```
lbesnard, line 88: same as comment above

AutomaticQC/SpikeClassifiers/imosSpikeClassifiersLimits.m
```diff
@@ -0,0 +1,119 @@
...
+%
+% limits - A structure with the limits.
+%
+% Example:
+%
+% len = 100
+% [limits] = imosTimeSeriesSpikeQCLimits(len)
+% assert(limits.hampel.window.min==1)
+% assert(limits.hampel.window.max=len/2)
+% assert(limits.otsu_savgol.pdeg.min=2)
```
lbesnard, line 20: this doesn't seem to match the structure below 
```
limits.hampel_half_window_width.min 
limits.otsu_savgol_pdeg.min
...
```

AutomaticQC/SpikeClassifiers/imosSpikeClassifiersLimits.m
```diff
@@ -0,0 +1,119 @@
...
+limits.otsu_threshold_nbins.min = int64(2);
+limits.otsu_threshold_nbins.max = int64(256);
+limits.otsu_threshold_nbins.help = '[int] The number of histogram bins when tracing the noise level. A larger value reduce global spike magnitude resolution.';
+
+limits.otsu_threshold_oscale.min = 1e-5;
+limits.otsu_threshold_oscale.max = 10;
+limits.otsu_threshold_oscale.help = '[float] A scale factor to multiply the noise threshold computed on the noise level. A large value increase the threshold (reduce detection).';
+
+limits.otsu_threshold_centralise.min = false;
+limits.otsu_threshold_centralise.max = true;
```
lbesnard, line 80: why are these booleans in this part? is it just to have 0 and 1 ? why not int64(false) ?

AutomaticQC/SpikeClassifiers/imosSpikeClassifierHampel.m
```diff
@@ -1,30 +1,25 @@
-function [spikes, fsignal] = imosSpikeClassifierHampel(signal, window, stdfactor)
-% function spikes, fsignal = imosSpikeClassifierHampel(signal, window, stdfactor)
+function [spikes, fsignal] = imosSpikeClassifierHampel(signal, half_window_width, madfactor, lower_mad_limit)
+% function spikes, fsignal = imosSpikeClassifierHampel(signal, half_window_width, madfactor)
```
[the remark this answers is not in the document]
ocehugo: ops, my bad -thanks

AutomaticQC/SpikeClassifiers/imosSpikeClassifierHampel.m
```diff
@@ -44,14 +39,20 @@
...
+if nargin == 1
+    half_window_width = 3;
+    madfactor = 10;
+    lower_mad_limit = 0.0;
+elseif nargin == 2
+    madfactor = 10;
+    lower_mad_limit = 0.0;
+elseif nargin == 3
+    lower_mad_limit = 0.0;
 end
```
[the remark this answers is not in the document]
ocehugo: no. although mad is never negative, a negative value input would behave like 0 here

AutomaticQC/SpikeClassifiers/imosSpikeClassifierNonBurstSavGolOTSU.m
```diff
@@ -51,47 +51,49 @@
...
 
-if window <= 3
-    error('Window need to be even and larger than 3')
-else
-
-    if rem(window, 2) == 0
-        window = window - 1;
-    end
+if window <= 2
+    error('Window need to be larger than 2')
```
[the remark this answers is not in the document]
ocehugo: > again where are these values found? is it just through testing?

The handling here of the window is a bit funky because SavGol filter only accepts odd windows. The defaults were set by a mix of "minimal effect" and "usability". 

In fact, the default at the end of this PR is 5 to be consistent with the behaviour of the filter. A window==3 actually did not change the value because of the implementation.

AutomaticQC/SpikeClassifiers/imosSpikeClassifierNonBurstSavGolOTSU.m
```diff
@@ -51,47 +51,49 @@
...
+else
+    fsignal = SavGol(signal, window/2, window/2, pdeg);
 end
 
-fsignal = SavGol(signal, (window - 1) / 2, (window - 1) / 2, pdeg);
 noise = signal - fsignal;
 abs_noise = abs(noise);
 otsu = otsu_threshold(abs_noise, nbins);
-threshold = otsu / oscale;
+threshold = otsu .* oscale;
```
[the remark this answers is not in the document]
ocehugo: Mostly for consistency with other methods mostly and docstrings. If you consider that the docstring was not on par with what the code was doing, yes, a bug!

Long story: In the original function, the threshold was arbitrary divided by 5. I just thought: "OK one more free parameter" and implement it as is (dividing), but the other implementations all work with multiplications on the scale, so this was odd behaviour compared to other methods and lead to confusion.

AutomaticQC/SpikeClassifiers/imosSpikeClassifiersLimits.m
ocehugo: yeap, another docstring wrong

AutomaticQC/SpikeClassifiers/imosSpikeClassifiersLimits.m
ocehugo: I handle specific types!That's why

AutomaticQC/SpikeClassifiers/imosSpikeClassifierBurstHampel.m
```diff
@@ -0,0 +1,176 @@
...
+% signal - A Burst 1-d signal.
+% use_burst_window  - a boolean to consider the half_window_width applied at burst scale. If false, will consider the burst series as continuous.
+% half_window_width - The half_window_width burst range (the full window size will be 1+2*half_window_width)
+% madfactor -  A multipling scale for the MAD.
+% repeated_only - a boolean to mark spikes in burst only if they are detected more than one time (only for half_window_width>0).
+% lower_mad_limit - a lower threshold for the MAD values, which values below will be ignored.
+%
+% Outputs:
+%
+% spikes - An array with spikes indexes.
```
lbesnard, line 19: when i look at the example and implementation, to me spikes is an array of spike values from the signal, not indexes. Am i missing something out?

AutomaticQC/SpikeClassifiers/imosSpikeClassifierBurstRunningStats.m
```diff
@@ -0,0 +1,74 @@
...
+% scalefun - the function to estimate the centre of the distribution. (e.g. mean or median)
+%          Default: @mean
+% dispersionfun - the function to estimate the dispersion (e.g. std, mean_absolute_deviation, median_absolute_deviation)
+%               Default: @std
+% dispersion_factor - the multiplying factor for the dispersion.
+%                   Default: 2
+%
+% Outputs:
+%
+% spikes - An array with spikes indexes.
```
lbesnard, line 20: aren't then booleans in that case

AutomaticQC/SpikeClassifiers/imosSpikeClassifierBurstHampel.m
ocehugo: the spikes output are always indexes where spike occur, not the actual value of the spike. You should not find anywhere the [spikes] output returning values, only indexes.

AutomaticQC/SpikeClassifiers/imosSpikeClassifierBurstRunningStats.m
ocehugo: no. always indexes.

ocehugo: Ow, I just realize there is a caveat here: I initialize spikes as a NaN double, but this is just for allocation and speed.
This array is filled with the  index values where spikes are found in the series, and them I just reduce the results to non-nans and unique values.

ocehugo: it's a trick for speed (avoid growing the array) and lowering the memory requirements - in matlab every cast to a variable is a copy, unless some very strict circustances are followed

AutomaticQC/SpikeClassifiers/imosSpikeClassifierBurstHampel.m
lbesnard: ok i really don't understand what is happening then

you have in your example
fullburst_spiked = [3,7,13,14,15,16,17,18,33,92,99]
...
assert(isequal(dspikes,[3,7,33,92,99]));

to me the values in the assert equal are pretty similar to the fullburst_spiked array, and not indexes

ocehugo: fullburst_spiked defined above is an array of indexes where the spikes are found. Just below this definition we got:
```
x(fullburst_spiked) = 1000;
```
Hence, these are the index locatiosn where values are 1000 in the random vector x.

The assertion you mention is correct. The function is configured to use the entire burst as a window. hence, It's missing the ```13,14,15,16,17,18``` because these are *all* the positions of a single burst (burst length here in the example is equal to 6). Hence, the entire burst is marked with value 1000. Thus, as configure per defaults, the burst statistics are mean/median == 1000 and std ==0; no spikes will be detected by definition. However, just below, if one seeks to detect those "entire bursts as a spike", you need to change the args to use more than one burst within a window. When this is the case, an entire burst with spiked data can be detected as per docstring.

lbesnard:
@ocehugo feel free to merge. Saying i understood it all would be a big lie but I tried!

AutomaticQC/SpikeClassifiers/imosSpikeClassifierNonBurstSavGolOTSU.m
```diff
@@ -51,47 +51,49 @@
...
     oscale = 1; %arbitrary scale used to reduce the otsu threshold
 elseif nargin == 2
     pdeg = 2; %quadratic polynomial
-    nbins = 256; % default bins in despiking1
+    nbins = 100; % default bins in despiking1
     oscale = 1; %arbitrary scale used to reduce the otsu threshold
 elseif nargin == 3
-    nbins = 256; % default bins in despiking1
+    nbins = 100; % default bins in despiking1
     oscale = 1; %arbitrary scale used to reduce the otsu threshold
```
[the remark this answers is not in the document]
ocehugo: This changes the values back to what was used in the original version of the code. It's also better because the defaults now are towards the limit range of the values.

ocehugo:
@lbesnard thanks - I will just fix the docstrings and merge

ocehugo:
ping @lbesnard - just need another aproval

PR #678, opened 2020-07-27
feat(bugfix): correct ADCP tilt flag broadcasting, fixes #680

### LINKED ISSUE #680: imosTiltVelocitytiltQC broken for ADCP
Tim sent me an email with TiltQC marking way more data than it should for ADCP.

The problem was introduced in 2.6.7 with code for current meters. The logic that handles the ADCP multidimensional array is incorrect (reversed). The problem was that the tests for ADCP were also incorrect with wrong dimensions in ROLL/PITCH, which allow this problem to pass unnoticed.

- [x] logic is corrected
- [x] tests are updated

ocehugo:
When adding support to current meters in TiltQC,
I implemented wrong assignment logic to the flags
that broke ADCP processing (multi dimensional flags).

The problem was not picked up by adcp testing because
those tests were also wrongly based on current meter templates,
and the pitch/roll dimensions were wrongly assigned and bypassed
proper evaluations.

### DIFF
--- a/AutomaticQC/imosTiltVelocitySetQC.m
+++ b/AutomaticQC/imosTiltVelocitySetQC.m
@@ -212,19 +212,18 @@ flags = ones(sizeCur, 'int8')*secondFlagThreshold;
 
 % tilt test
 iPass = tilt < secondTiltThreshold;
-if ~isvector(flags)
+if isvector(flags)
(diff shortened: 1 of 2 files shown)


ocehugo:
ping @lbesnard

PR #681, opened 2020-08-14
Fix fieldname access imosTimeSeriesSpikeQC.m

sspagnol:
Testing for sample_data.FIELDNAME but accessing sample_data.meta.FIELDNAME


ocehugo:
@sspagnol , is this blocking anything important apart from netcdf?

sspagnol:
For me imosTimeSeriesSpikeQC.m was run when it was in the list of QC data routines, then matlab would throw error and so wouldn't get to the spike window (the one with list of variable and preview button). Dont' know about netcdf stuff. And only tested on a dataset the Vinnie was having trouble with.

ocehugo:
Thanks. Just reproduced.

Lack of automated tests with real files on that one. damn, just minted a release...will need to do another and reverse the check here.

ocehugo:
fixed in #682

PR #683, opened 2020-08-25
675 rebase

ocehugo:
@BecCowley - this is the bare bones I want to develop on.

What I did:

a. Dropped some commits with templating and local options that are not required.
b. Squashed some changes in certain files
c. Split some big commits into several, so we have some functionality closure.
d. rewording commits
e. Reordered (for further rebase/fixups if necessary).

I put the order of the commits in roughly the order I want to address things, so it's easy to rebase further if required. However, what I need now is to reach a self-contained test and reproduce the results. After that, I can go and do further cleanups/changes/etc.

Testing:

1. I can't properly execute your test with the files you sent me. The file 17959001.000 is too big and too slow to load. After reducing the binary file I could import it but then you call the UI for selecting the pre-processing, and I'm not sure which ones you use (since toolboxproperties changes a lot). Could you confirm if the toolboxproperties you sent on the previous PR is the one you use for the tests? Otherwise, just a list of the PP used is fine for me.

2. As well, I just need some basic metadata from 17959001.000 file (site depth, nominal depths and such). Would you please send me the sample_data struct so I can copy the metadata? This will also help in a self-contained test and no need for dbs.

3. The same applied for the thresholds, Which one you selected for this file? We will improve the threshold plots later, all i want is to have a robust auto-test so I can start the cleanup process. The same applied to some parameters you define in SurfaceDetectionSet (the PR doesn't have one, and we need static values for the test).

4. I kept the Geomag changes, but are they required for 17959001.000? Any other test case will require these fields?

Thanks


BecCowley:
@ocehugo 
1. The files are huge, that is part of the problem! I have to run all the processing on a unix box and that is very slow. I don't think it is possible to run on a standard desktop PC. Yes, the toolboxproperties file is what I am using. The PP tests I use are timeDrift, DepthPP and magnetic declination. I am happy to go through the process of how I use the single_pingADCPtest.m code if you want. I call the autoQC steps one at a time to determine what the best thresholds are, so I run that section of the code at least once for each threshold test. It also requires deleting the *.pqc file each time so that the values I select are used, not the ones in the *.pqc file.

2. I will send the sample_data struct. If you want the original csv files ( I don't use a database), I can send that too.

3. Thresholds for this file. Echo Intensity and SideLobe thresholds are used as part of the 'SurfaceDetection' test I wrote. I don't use the individual EchoIntensity and SideLobe tests.

Echo Range threshold | Echo Intensity threshold | SideLobe threshold | Cmag threshold | Error Velocity threshold | HV threshold | VV threshold | Tilt flag 2 | Tilt flag 4
-- | -- | -- | -- | -- | -- | -- | -- | --
55 | 30 | 0.5 | 100 | 0.4 | 2.5 | 1.2 | 40 | 60


4. I'm not sure what changes I made to the geomag, except maybe they are the user settings that are recorded when the routine is called?

ocehugo:
@BecCowley 
>1. The files are huge, that is part of the problem! I have to run all the processing on a unix box and that is very slow. I don't think it is possible to run on a standard desktop PC.

The file format clearly wastes a lot of space, but the actual problem is that the parser is too slow and memory hungry. Since I need something that will run in under seconds from raw file to the end of parsing (test), I reduced the binary to the first 100mb, which creates a sample_data with roughly 50mb of valid data, if my quick inspection was correct. See the file [here](https://universitytasmania-my.sharepoint.com/:u:/g/personal/hugo_oliveira_utas_edu_au/EaB5Wj98279EgcD5du9j3CQBZgGXE27KxsrRiClwfPTypw?e=xop16E)

I'm just not sure if this file will suffice for all the PR/workflow checks - let me know - but the above file is what I got in mind to use. 

> Yes, the toolboxproperties file is what I am using. The PP tests I use are timeDrift, DepthPP and magnetic declination. I am happy to go through the process of how I use the single_pingADCPtest.m code if you want.

Great. It's a good idea to see you doing it so I can change the current test from interactive to explicit code/function calls. Ping me over email with the time you wanna do it, I'm quite free today/this week.

> I call the autoQC steps one at a time to determine what the best thresholds are, so I run that section of the code at least once for each threshold test. It also requires deleting the *.pqc file each time so that the values I select are used, not the ones in the *.pqc file.

OK. By calling explicitly the routines we avoid the pqc problem. I think I will understand better once you demoed to me. Seems to me that we will need a primary QC routine that will execute secondary ones to aid in the decision, but this is for later anyway.

> 2. I will send the sample_data struct. If you want the original csv files ( I don't use a database), I can send that too.
> 3. Thresholds for this file. Echo Intensity and SideLobe thresholds are used as part of the 'SurfaceDetection' test I wrote. I don't use the individual EchoIntensity and SideLobe tests.
Echo Range threshold | Echo Intensity threshold | SideLobe threshold | Cmag threshold | Error Velocity threshold | HV threshold | VV threshold | Tilt flag 2 | Tilt flag 4
-- | -- | -- | -- | -- | -- | -- | -- | --
55 | 30 | 0.5 | 100 | 0.4 | 2.5 | 1.2 | 40 | 60

Thanks!

>4. I'm not sure what changes I made to the geomag, except maybe they are the user settings that are recorded when the routine is called?

OK, it's just dynamic IO then and we don't care about it. Will drop the commit when the test is refactored.

AutomaticQC/imosEchoIntensityVelocitySetQC.m
```diff
@@ -113,29 +117,66 @@
 % same flags are given to any variable
 flags = ones(sizeCur, 'int8')*rawFlag;
 
+isUpwardLooking = true;
+if all(Bins <= 0), isUpwardLooking = false; end
```
sspagnol, line 121: Just checking if sample_data.dimensions{idHeight}.data only contains in water data? Just comparing parsers (say line 315 of workhorseParse.m)
```
adcpOrientation = mode(adcpOrientations); % hopefully the most frequent value reflects the orientation when deployed
```

And following on from that, if the parsers have already calculated orientation once (which it looks like aquadoppProfiler, awac, contental, signature, workhorse do) then can't this be saved in the sample_data.meta area for reuse.

Parser/workhorseParse.m
```diff
@@ -167,13 +168,42 @@
...
+%       if adcpOrientation == 1
+%           adcpOrientation = 0;
+%           adcpOrientations(adcpOrientations==1)=0;
+%       else
+%           adcpOrientation = 1;
+%           adcpOrientations(adcpOrientations==0)=1;
+%       end
+%   end
+      
+  if mode(fixed.coordinateTransform) == 31 %ENU
```
sspagnol, line 192: Regards coordinate system type (ENU,BEAM), In local code for modified workhorseParse.m have comment/code 
```
  % EX/coordinate transform
  % xxx00xxx = NO TRANSFORMATION (BEAM COORDINATES)
  % xxx01xxx = INSTRUMENT COORDINATES
  % xxx10xxx = SHIP COORDINATES
  % xxx11xxx = EARTH COORDINATES
  % xxxxx1xx = TILTS (PITCH AND ROLL) USED IN SHIP OR EARTH TRANSFORMATION
  % xxxxxx1x = 3-BEAM SOLUTION USED IF ONE BEAM IS BELOW THE CORRELATION THRESHOLD SET BY THE WC command
  % xxxxxxx1 = BIN MAPPING USED
  sect.coordinateTransform = double(data(idx+25));
  cfg.coordinateSystem     = getopt(bitand(bitshift(mode(sect.coordinateTransform),-3),3),'BEAM','INSTRUMENT','SHIP','EARTH');
```
The idea from R. Pawlowicz rdradcp.m

Just want to highlight that the 7/31 magic numbers might be dependent on other bits set in ```sect.coordinateTransform```

And if all adcp's have at least the concept of BEAM and EARTH coordinates, then this configuration could be determined in the parsers and stored in sample_data.meta somehow (just equivalently amongst all adcp parsers).

AutomaticQC/imosEchoIntensityVelocitySetQC.m
ocehugo: yes, this should be within the meta struct. I would avoid creating more reviews here @sspagnol because I can still rebase this a bit and the comments be lost.

ocehugo: to make sure, just create further comments at the top-level PR page!

BecCowley:
@ocehugo I loaded the reduced size file and it tested fine. Just under half of it is out-of-water data. If you are happy with that, then it's good to test on.

Parser/rdiBeam2Earth.m
```diff
@@ -0,0 +1,194 @@
+function [veast,vnrth,wvel,evel]=rdiBeam2Earth(b1,b2,b3,b4,head,pitch,roll,ba,cvxccv,updown, distance)
+% Function to convert ADCP beam velocities to Earth coordinate velocity.
+% b1,b2,b3,b4 are four beam velocities
+% head is the heading reading, in degrees.
```
ocehugo, line 4: AFAIK the inputs here are not in degrees, since the call to this function is before converting the raw readings to degrees (LSD=0.01 in the manual).

Parser/rdiBeam2Earth.m
```diff
@@ -0,0 +1,194 @@
...
+% cvxccv is 1 for a convex xdcr, other than 1 for a concave xdcr.  Usually 1.
+% updwon is 0 for a down-looking instrument, other than 0 for an up-looker.
+%
+% Calls matlab routines beam2inst.m and inst2earth.m.
+%
+head = head*0.01;
+pitch = pitch *0.01;
+roll = roll * 0.01;
+% make pitch, roll, heading the same size as b1, b2, b3, b4
+elev=[-70 -70 -70 -70]; % default elev for 20o
```
ocehugo, line 17: dead code

Parser/rdiBeam2Earth.m
```diff
@@ -0,0 +1,194 @@
...
+if updown == 1 %uplooking
+    txtup = 'up';
+else
+    txtup = 'down';
+end
+[veast,vnrth,wvel,evel] = deal(NaN*b1);
+% have to bin map it first. Even if the EX bit is set for bin
+% mapping, if the data is collected in beam coordinates, bin mapping
+% does not occur on board.
+[b1,b2,b3,b4] = binmap(b1,b2,b3,b4,head,pitch,roll,ba,distance);
```
ocehugo, line 28: better to stay with a single unit in the entire functionality. There are double conversion steps all over the place. Better to stick with radians from the start.

Parser/rdiBeam2Earth.m
```diff
@@ -0,0 +1,194 @@
...
+function [b1m,b2m,b3m,b4m] = binmap(b1,b2,b3,b4,head,pitch,roll,ba,distance)
+distAlongBeams = distance';
+pitch = pitch*pi/180;
+roll  = roll*pi/180;
+
+beamAngle = ba*pi/180;
+nBins = length(distAlongBeams);
+
+% RDI 4 beams
+nonMappedHeightAboveSensorBeam1 = (cos(beamAngle + roll)/cos(beamAngle)) * distAlongBeams;
```
ocehugo, line 123: should be a function

Parser/rdiBeam2Earth.m
```diff
@@ -0,0 +1,194 @@
...
+    rrad=roll*pi/180;
+else
+    rrad=(roll+180)*pi/180;
+end
+zero= 0;
+one = 1;
+%
+% Create the heading rotation matrix:
+%
+h=[cos(hrad)  sin(hrad)  zero zero
```
ocehugo, line 65: I would prefer to write the entire matrix here instead. I'm also wondering if we need to handle ship roll  (any case for that) and the use tilt ==0 in the definition of this matrix

Parser/rdiBeam2Earth.m
```diff
@@ -0,0 +1,194 @@
...
+    txtup = 'down';
+end
+[veast,vnrth,wvel,evel] = deal(NaN*b1);
+% have to bin map it first. Even if the EX bit is set for bin
+% mapping, if the data is collected in beam coordinates, bin mapping
+% does not occur on board.
+[b1,b2,b3,b4] = binmap(b1,b2,b3,b4,head,pitch,roll,ba,distance);
+for a = 1:length(pitch)
+    beamdat = [b1(a,:); b2(a,:); b3(a,:); b4(a,:)];
+    thePitch=atand(tand(pitch(a))*cosd(roll(a))); % CBluteau as per RDI coordinate transformation manual Eq 19 when EZxxx1xxx is set
```
ocehugo, line 31: need to check/pass the EZxxx1xxx argument to this function.

Parser/rdiBeam2Earth.m
```diff
@@ -0,0 +1,194 @@
...
+% instrument coordinates based the beam angle.
+%
+% ba is the beam angle of the ADCP (typically either 20 or 30 degrees)
+% c is whether the instrument is convex (c=1) or concave (c other than 1). Nearly
+% all ADCPs are convex (some early vessel-mounted systems were concavbe).
+%
+% Create the apropriate scale factors:
+%
+barad=ba*pi/180;
+a=1/(2*sin(barad));
```
ocehugo, line 101: same as above, lets use the same units in the functionality and avoid unecessary operations.

ocehugo: I also wonder if we could use symbolic maths here for more concise execution.

Parser/rdiBeam2Earth.m
```diff
@@ -0,0 +1,194 @@
...
+nonMappedHeightAboveSensorBeam4 = repmat(cos(roll), 1, nBins) .* nonMappedHeightAboveSensorBeam4;
+
+nSamples = length(pitch);
+mappedHeightAboveSensor = repmat(distAlongBeams, nSamples, 1);
+
+% we can now interpolate mapped values per bin when needed for each
+% impacted parameter
+
+% only process variables that are in beam coordinates
+for beamnumber = 1:4
```
ocehugo, line 142: this should be a function too

Parser/workhorseParse.m
```diff
@@ -167,13 +168,42 @@
...
+  % check for ENU and Beam coordinates. Instrument coordinates not
+  % considered.
   %
-  
-  veast = velocity.velocity1;
-  vnrth = velocity.velocity2;
-  wvel  = velocity.velocity3;
-  evel  = velocity.velocity4;
+  adcpOrientations = str2num(fixed.systemConfiguration(:, 1)); % str2num is actually more relevant than str2double here
+  adcpOrientation = mode(adcpOrientations); % hopefully the most frequent value reflects the orientation when deployed
```
ocehugo, line 175: this value should be exported forward in meta.

Parser/workhorseParse.m
ocehugo: the mode should be forwarded in meta. The parser should not do the conversion - this is the work of PP routines.

Parser/workhorseParse.m
```diff
@@ -167,13 +168,42 @@
...
+      
+  if mode(fixed.coordinateTransform) == 31 %ENU
+      veast = velocity.velocity1;
+      vnrth = velocity.velocity2;
+      wvel  = velocity.velocity3;
+      evel  = velocity.velocity4;
+  elseif mode(fixed.coordinateTransform) == 7 %Beam coordinates
+      %convert to earth coordinates
+      %concave/convex:
+      cvxccv = mode(str2num(fixed.systemConfiguration(:,5)));
```
ocehugo, line 200: is_concave/is_convex is much cleaner

ocehugo (commented):
@BecCowley , may you comment below your thoughts!? I'm towards the end of the revamp for this commit, so would appreciate some feedback on the comments below.

Parser/workhorseParse.m
BecCowley: Yes, agree.

Parser/workhorseParse.m
BecCowley: Yes, agree. 
Also, there is a bit of an issue with using the mode and hoping it is correct. If the record is very short (due to a failure), and the majority of the data is 'out of water' and facing the other way, then the orientation will be identified incorrectly. There is no way for the user to put in their own selection to override the automatic choice. Maybe this is a new issue.

Parser/rdiBeam2Earth.m
BecCowley: This code came from someone else (I think the Scripps group). Possibly their input was in degrees. I hope that I was using degrees to input. Need to check what the parser does (does it do a change when reading into the sample_data structure)?

Parser/rdiBeam2Earth.m
BecCowley: Yes, whatever is easiest. As long as in the end, we can see the heading in degrees, I think.

Parser/rdiBeam2Earth.m
BecCowley: This might need some more investigation to compare to what the RDI tools do, but I think when I ran this as is, it gave the same output as the RDI Velocity software. But, I am only looking at moored instruments, not ship-mounted ADCPs.

Parser/rdiBeam2Earth.m
ocehugo: all is fine, it was just a remark because the help is misleading. There is a conversion just below (*0.01), which puzzled me,  until I went to the workhorse manual and found this LSD factor. Which also lead me to think in the possibility of double conversions cases. These are the oddities I need to be careful about since I'm shifting the code calls to do things more in line with what we do in other places.

Parser/rdiBeam2Earth.m
ocehugo: OK, I will fence this to stop if we ever face a ship-mounted adcp stream. I just need to find this in the parse and handle to this (another argument).

PR #685, opened 2020-09-24
feat(UI): detailLevel and aliases for instruments

ocehugo:
These changes permit the user to set
the level of detail shown on all UI elements related
to instrument metadata description.

Two new options are provided in the toolboxProperties,
`instrumentAliases` and `detailLevel`.

instrumentAliases modify the display name of instruments
in the UI to the entries presented within
the GUI/instrumentAliases.txt file.

detailLevel modify the data description in all UI elements
to match several named options.

See the updated toolboxProperties.txt file for details.

### DIFF
--- a/Util/readMappings.m
+++ b/Util/readMappings.m
@@ -1,0 +1,49 @@
+function [mappings] = readMappings(file, delimiter)
+% function [mappings] = readMappings(file, delimiter)
+%
+% This read a mapping file, with a predefined delimiter (',').
+%
+% Inputs:
+%
+% file - a file location string
+% delimiter - a field delimiter
+%             default: ','
+%
+% Outputs:
+%
+% mappings - a containers.Map mapping
+%
+% Example:
+%
+% file =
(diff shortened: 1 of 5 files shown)


Util/readMappings.m
```diff
@@ -0,0 +1,49 @@
...
+%
+% Outputs:
+%
+% mappings - a containers.Map mapping
+%
+% Example:
+%
+% file =
+% [mappings] = readMappings(file)
+% assert()
```
lbesnard, line 20: would be good to write the assert. also, no need for a unittest?

lbesnard (dismissed):

Util/readMappings.m
ocehugo: Yes, Good catch. I fixed some aspects of the func days ago because a bug bite me just because of that missing docstring unittest.

I forgot to push the commit here though. See new commit below.

lbesnard (approved):

PR #686, opened 2020-09-25
some new test utilities 

ocehugo:
These are some utilities to reduce some manual testing, code size in test classes, and also to avoid write test classes in some cases.

`test_docstring` enable us to use the function docstring as tests. It reads the docstring blocks within the `%Example:`
subblock and evaluate every expression. It's simple (no multiline evaluation), but help us to reduce code size and make sure docstrings are updated.

EDIT: moved some commits to another branch to be self contained.


ocehugo:
@lbesnard - I just rebased to be more self-contained and given some related changes down the road (since no review was done here anyway).

lbesnard:
some of these functions are already in the 687 PR

ocehugo:
Yes, they were nested/dependent.

PR #687, opened 2020-10-26
Docstring fixes and small refactorings

ocehugo:
This branch contains some fixes triggered by #686 functionality.

### DIFF
--- a/Util/File/read_until_match.m
+++ b/Util/File/read_until_match.m
@@ -1,0 +1,109 @@
+function [clines, number_of_lines] = read_until_match(fid, pattern, is_regex, preallocsize)
+%function [clines, number_of_lines] = read_until_match(fid, pattern, is_regex, preallocsize)
+%
+% Read a file, line by line, until a pattern is found, returning
+% all lines read and the total number of lines.
+% The result is empty (total number of lines) if no match is found.
+%
+% Inputs:
+%
+% fid - numerical file id
+% pattern - the string pattern to match
+% is_regex - the string pattern is a regex.
+%            Default: False.
+% preallocsize - the preallocation length (N) of the clines cell (1xN).
+%
+%
+% Outputs:
+%
+% clines - a cell of strings with all lines read until the match.
+% number_of_lines - the total number of lines read
+%
+% Example:
+% % create
+% tmpfile = [tempdir 'test_read_until_match.txt'];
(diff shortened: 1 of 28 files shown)


Util/File/read_until_match.m
```diff
@@ -0,0 +1,109 @@
+function [clines, number_of_lines] = read_until_match(fid, pattern, is_regex, preallocsize)
```
lbesnard, line 1: function name doesn't match convention

Util/Random/random_between.m
```diff
@@ -0,0 +1,53 @@
+function [arr] = random_between(a,b,n,type)
```
lbesnard, line 1: same as above and other function names in this PR. but why those "random" functions? where are they used?

Util/Random/random_between.m
```diff
@@ -0,0 +1,53 @@
...
+%
+% % random constant
+% arr=random_between(1,1,10,'int');
+% assert(isequal(arr,ones(1,10)));
+%
+%
+% author: hugo.oliveira@utas.edu.au
+%
+
+if nargin>3
```
lbesnard, line 39: more specific is n=4, but doesn't really matter

lbesnard (changes requested):

Util/File/read_until_match.m
ocehugo: Yeah, the reason why was that these were all internal functions. However, I would like to avoid changing them now and have to rebase 3 other branches that are build on top of these commits.
Moreover, I will need to improve `test_docstring` anyway (support evaluation of docstring examples in classes), so these can go together later.

ocehugo: or I can do a isolated commit on top of the last branch to make things easier.

Util/Random/random_between.m
ocehugo: as above for renaming. I used them on some unpushed tests and for automatic generation of some things. Typically, these include checking randomly cell indexes.

Util/Random/random_between.m
ocehugo: Yeap.

ocehugo:
@lbesnard - points addressed and very small things that I noticed when renaming them.

lbesnard (approved):

PR #725, opened 2021-02-23
Workhorse parse overhaul

ocehugo:
This is Part 1 out of 3 of the next Toolbox release.

See commit msgs from Jan 15 onwards, since changes were built on top of  https://github.com/aodn/imos-toolbox/pull/683.

@BecCowley - This is not complete, but you could just check if we are parsing correctly your new files. Hopefully tomorrow I can push part 2 (PP conversions), followed by the QC routines and the actual release (part3).


ocehugo:
As always, at the end of the day, I forgot to re-run all the tests before sending this, so had to fix some misc things. It is now ready.

Parser/+Workhorse/translate_system_configuration.m
```diff
@@ -0,0 +1,145 @@
...
+% sysconfig[logical] - The system configuration logical array representing the 8 LSB
+%                      bits and the 8 MSB bits.
+%                      Note: Bit-padding is assumed (length(sysconfig)==16)
+%
+% Outputs:
+%
+%  adcp_system - A struct with all derived system config.
+%             LSB derived fields:
+%
+%  adcp_system.beam_face_config[char] - Static beam facing config [ 'up' | 'down' ]
```
lbesnard, line 22: same comment as above. Could this be a side config as well?

Parser/+Workhorse/translate_system_configuration.m
```diff
@@ -0,0 +1,145 @@
...
+        adcp_system.xmit_voltage_scale = 380667;
+    case '100'
+        adcp_system.system_freq = 1200;
+        adcp_system.model_name = 'Sentinel or Monitor';
+        adcp_system.xmit_voltage_scale = 253765;
+    case '101'
+        adcp_system.system_freq = 2400;
+        adcp_system.model_name = 'DVS';
+        adcp_system.xmit_voltage_scale = 253765;
+    otherwise
```
lbesnard, line 113: wouldn't this be better to be in a config file? same with a few other things below. This would avoid writing all the case functions

Parser/+Workhorse/cell_cdistance.m
```diff
@@ -0,0 +1,46 @@
...
+% assert(d(3)==first_bin_cdist+2*cell_length)
+% assert(d(end)==first_bin_cdist+(num_cells-1)*cell_length)
+%
+% % Down facing is negative
+% d = Workhorse.cell_cdistance(first_bin_cdist,cell_length,num_cells,'down');
+% assert(all(sign(d)==-1))
+%
+narginchk(4, 4)
+
+if strcmpi(beam_face_config, 'down')
```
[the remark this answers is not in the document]
ocehugo: This beam_face_config is actually a defined option/parameter in the RDI software output. As discussed, we don't handle them (yet!?), and a side looking adcp will probably have a different instrument entry (not workhorse).

Parser/+Workhorse/conversion_mappings.m
```diff
@@ -0,0 +1,92 @@
+function [cmap] = conversion_mappings(sensors, name_extension, xmit_voltage_scale, frame_of_reference)
+%function [cmap] = conversion_mappings(sensors, name_extension,xmit_voltage_scale, frame_of_reference)
+%
+% Load the Conversion mappings for workhorse ADCP
+%
+% Inputs:
+%
+% sensors [struct] - A structure with sensors switches.
+% name_extension [char] - The variable anme extension for velocity and Heading.
```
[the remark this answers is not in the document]
ocehugo: done

Parser/+Workhorse/translate_system_configuration.m
ocehugo: these values are pretty much static. We can't actually write nice structs in config files, unless we go to json/toml which is an overkill here.

lbesnard (approved):

PR #730, opened 2021-03-09
Rdi bin mapping and rotations, Part 2.

ocehugo:
This is the userland RDI ADCP bin mapping and beam to earth rotation routines. 

See commit msgs and regression/unit tests.


ocehugo:
@lbesnard - I just relink the PR and the merged commits are gone.

lbesnard (approved):

PR #731, opened 2021-03-28
Minor improvements

BecCowley:
Update to sites.txt to add EAC1520 site (previously EAC1 in the 2012-2013 deployment. Also, allow resizing of the depthPP window to enable me to read the full names of the instruments.


BecCowley:
Probably a lot of merging stuff that you don't need. Just these two: 1de988f and 8126380

PR #754, opened 2021-09-24
Bug binmapping

BecCowley:
This fix is a sign adjustment to the bin mapping algorithm based on the direction the instrument is facing. Current code is only correct for down-facing ADCPs.


BecCowley:
Closing, missing some changes, will re-submit merge request.

PR #756, opened 2021-09-24
Bug error velocity QC test

BecCowley:
Should have an & to find error velocity values that are between the two extremes.


evacougnon:
@BecCowley , I'm creating a new PR with your work to only include the last commit `e1cc8ca`.

Closing this PR as the specific commit is now in #767

PR #759, opened 2021-09-30
(Fix) Replace references of old facility to DWM

lbesnard:


test/Parser/testnetcdfParse.m
```diff
@@ -2,7 +2,7 @@
 
     properties (TestParameter)
         anmn_file = files2namestruct(only_netcdf(rdir([toolboxRootPath 'data/testfiles/aodn/ANMN'])))
-        abos_file = files2namestruct(only_netcdf(rdir([toolboxRootPath 'data/testfiles/aodn/ABOS'])))
+        dwm_file = files2namestruct(only_netcdf(rdir([toolboxRootPath 'data/testfiles/aodn/DWM'])))
```
ggalibert, line 5: you will need to change the directory structure accordingly where the test files are stored.

test/Parser/testnetcdfParse.m
```diff
@@ -15,8 +15,8 @@ function test_no_error_reading_anmn(~, anmn_file)
             assert_burst_sampling(data);
         end
 
-        function test_no_error_reading_abos(~, abos_file)
-            data = netcdfParse({abos_file}, '');
+        function test_no_error_reading_dwm(~, dwm_file)
```
ggalibert, line 18: you will need to update this everywhere this function is called too.

test/Parser/testnetcdfParse.m
lbesnard: This is already done abd explained in details here:
https://github.com/aodn/PO-Backlog/issues/2185#issuecomment-930797090

test/Parser/testnetcdfParse.m
lbesnard: This function is not called "directly" by its name. It is called as part of the unittest suite. It's not "referenced anywhere else

Issue #791, opened 2022-08-03
does IMOS read AD2CP bottom-track incorrectly?


dankelley:
# Introduction

I have been working to add support for AD2CP data to the "oce" package in the R language.  I am not sure I am reading bottom-track data correctly (see e.g. https://github.com/dankelley/oce/issues/1982), and a websearch turned up the IMOS package.  Although I am not a matlab user anymore, I think I can understand your code, and I think there may be an error in it.

# Possible IMOS Error

At https://github.com/aodn/imos-toolbox/blob/e19c8c604cd062a7212cdedafe11436209336ba5/Parser/readAD2CPBinary.m#L558 I see that ambiguity velocity is being read as a two-byte (unint16) value.  However, my understanding is that it ought to be 4 bytes for bottom-track data (in contrast to other data types, which are 2 bytes).  If I am correct in my interpretation, then velocity, etc. may be inferred incorrectly by IMOS (see e.g. https://github.com/aodn/imos-toolbox/blob/e19c8c604cd062a7212cdedafe11436209336ba5/Parser/readAD2CPBinary.m#L569 for the offset used for velocity), and that is why I am writing today.  Hopefully, I am wrong, or at least if I'm right, perhaps this can lead to an adjustment to the IMOS code.

My sources for the 4-byte nature of ambiguity-velocity (for bottom-track, not other types) are as follows.

1. Page 61 of Nortek (2017), as shown in the first snapshot below.

2. Page 94 of Nortek (2022), as shown in the second snapshot below.

# References

* Nortek AS. “Signature Integration 55|250|500|1000kHz (2017).” Nortek AS,
  2017.
https://www.nortekgroup.com/assets/software/N3015-007-Integrators-Guide-AD2CP_1018.pdf.

* Nortek AS. “Signature Integration 55|250|500|1000kHz (2022).” Nortek AS,
  March 31, 2022.

# Snapshot 1 (2017 manual)

<img width="751" alt="N2017" src="https://user-images.githubusercontent.com/99469/182600775-016fa829-13e2-404d-9921-c82f341613bc.png">

# Snapshot 2 (2022 manual)

<img width="633" alt="N2022" src="https://user-images.githubusercontent.com/99469/182600981-68963cc4-fb8f-433d-9295-0206160dfd3d.png">

kereid:
[...]

PR #794, opened 2023-01-05
Fix(readAD2CPBinary): various issues with Bottom tracking parser

lbesnard:
* the previous version of the parser could not possibly work. variable names
     were not always used properly
 * amend the index values to read various Nortek variables to match the Nortek 2022 manual
 * add unittest with bottom tracking provided by Clark Richards

### DIFF
--- a/Parser/readAD2CPBinary.m
+++ b/Parser/readAD2CPBinary.m
@@ -513,12 +513,13 @@ end
 
 end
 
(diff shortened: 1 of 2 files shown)


lbesnard:
linked to https://github.com/aodn/imos-toolbox/issues/791

evacougnon (approved):

PR #797, opened 2023-06-27
(Feat) Add failed QC tests information into new ancillary variable [*]_failed_tests

lbesnard:


NetCDF/exportNetCDF.m
```diff
@@ -470,7 +512,7 @@
 %
   switch(type)
     case 'dimensions'
-      var = sample_data.dimensions{varIdx};
+      var = sample_data.dimensions{varIdx}; % when is this even triggered? (Loz, 20230624)
```
lbesnard, line 515: @ggalibert I wouldn't mind your input on this? I tried debugging and couldn't see this line being ever reached

FlowManager/displayManager.m
```diff
@@ -623,8 +623,20 @@ function resetPropQCCallback()
           
           for j=1:length(resetIdx)
               for i=1:length(sample_data{resetIdx(j)}.variables)
-                  if isfield(sample_data{resetIdx(j)}.variables(i), 'ancillary_comment')
-                      sample_data{resetIdx(j)}.variables(i) = rmfield(sample_data{resetIdx(j)}.variables(i), 'ancillary_comment');
+                  % previous code could not run since it was using variables(i) instead of variables{i}.
+                  % TODO ask @ggalibert about this
+                  if isfield(sample_data{resetIdx(j)}.variables{i}, 'ancillary_comment')
```
lbesnard, line 628: @ggalibert quick question about syntax. Previous code didn't seem to be able to run. Am I missing something re matlab syntax?

NetCDF/exportNetCDF.m
ggalibert: Probably never, if you look for where addQCvar function is called, you will see it is not called for dimensions.

FlowManager/displayManager.m
ggalibert: Must have been a typo, if sample_data.variable is a cell array then curly bracket should be used. Makes sense it is a cell array since not all variables structures have the same fields.

lbesnard:
Some test files generated with this draft PR:
https://cloudstor.aarnet.edu.au/plus/s/RKMb19mu8O6GSQx

mhidas:
:+1: Thanks @lbesnard 
Here's how I did the same thing...
```python
def map_failed_tests(var):
    """Convert a _failed_tests variable to a list of strings.
    var can be an xarray.DataArray or a netCDF4.Variable
    """

    flag_map = dict(zip(var.flag_masks, var.flag_meanings.split()))
    def map_one(value):    
        ms = [flag_map[i] for i in flag_map if value & i]
        return ", ".join(ms)

    return list(map_one(v) for v in var.data)
```

ggalibert (approved):

